PR #6, opened 2021-09-23
added contract_22_defect_check

### LINKED ISSUE #5: Add check for CSCvz65560
https://www.cisco.com/c/en/us/td/docs/switches/datacenter/aci/apic/sw/advisory/aci-field-notice-CSCvz65560.html

CSCvz65560 - add/update contract with TCP port 22 cause switch rule disable if APIC is 5.x but switch is 14.x

monrog2:
fixes #5


monrog2:
signed off by TK

PR #52, opened 2023-05-25
Make it compatible with both py27 and py38

monrog2:
Older APIC firmware versions only have py27.
Relatively older APIC firmware versions use py27 as the default. Newer APIC firmware versions use py38 as the default and no longer has py27 at all.
To support any APIC firmware version, make this script compatible with both py27 and py38.


monrog2:
pytest on py3 worked after adding `six` dependency.

tested on various pre 6.x fabrics with success.

merging.

PR #53, opened 2023-05-25
Docker0 overlap

monrog2:


monrog2:
Local test on APIC 3.2.7f fails on import:
```
admin@apic1:techsupport> python preupgradeChecks.py
Traceback (most recent call last):
  File "preupgradeChecks.py", line 10, in <module>
    from ipaddress import IPv4Address, IPv4Network
ImportError: No module named ipaddress
```

4.2 and 5.2 tests do not have that issue

PR #63, opened 2023-08-03
Sup hwrev check + v1.6.0 bump

monrog2:


monrog2:
pytest - passed
integration test - passed

PR #65, opened 2023-08-10
bugId to sup HWRev check and update README

monrog2:


monrog2:
pytests  passed
integration tests passed

PR #96, opened 2024-04-19
New validation to address CSCwj69435 for the afected versions

ehaminian:
Following a recent outage with one of our clients, we've noticed an issue with the APIC failing to reject configurations where "Access(untagged)" mode is applied to a single interface across multiple EPGs. This configuration oversight, likely due to user error, poses a significant risk. Any switch reprogramming trigger could inadvertently deploy new VLANs/EPGs on interfaces, a scenario that should be prevented by the APIC during configuration.

To address this issue, we've filled a defect under CSCwj69435. We've also introduced a new validation mechanism, titled "Overlapping Access (Untagged) mode on interfaces," which will detect and warn customers about this issue before any upgrades are performed.

This validation process involves checking individual interfaces/port-channels, VPC interfaces, and FEX interfaces using the three regex patterns defined in the code.


takishida (changes requested):
Thank you for the PR.

This is a good catch, but this check falls in to the same category as the following checks in the sense of fault F0467:

* `port_configured_for_apic_check`
* `port_configured_as_l2_check`
* `port_configured_as_l3_check`
* `prefix_already_in_use_check`
* `encap_already_in_use_check`

As you can see from the fault below, it's the same fault as the checks shown above, but with a different `configQual` in the `changeSet`. 

```
admin@apic1:~> moquery -c faultInst  -x 'query-target-filter=wcard(faultInst.changeSet,"native-or-untagged-encap-failure")'
Total Objects shown: 1

# fault.Inst
code             : F0467
ack              : no
alert            : no
annotation       :
cause            : configuration-failed
changeSet        : configQual (New: native-or-untagged-encap-failure), debugMessage (New: )
childAction      :
created          : 2024-04-19T17:57:59.887-07:00
delegated        : yes
descr            : Configuration failed for uni/tn-TK/ap-AP1/epg-EPG1-2 node 101 eth1/4 due to Only One Native or Untagged Encap Allowed on Interface, debug message:
dn               : topology/pod-1/node-101/local/svc-policyelem-id-0/uni/epp/fv-[uni/tn-TK/ap-AP1/epg-EPG1-2]/node-101/stpathatt-[eth1/4]/nwissues/fault-F0467
domain           : tenant
extMngdBy        : undefined
highestSeverity  : minor
lastTransition   : 2024-04-19T18:00:23.777-07:00
lc               : raised
modTs            : never
occur            : 1
origSeverity     : minor
prevSeverity     : minor
rn               : fault-F0467
rule             : fv-nw-issues-config-failed
severity         : minor
status           :
subject          : management
title            :
type             : config
uid              :
userdom          : all
```

For consistency and extensibility, we should follow the same approach as the other F0467 checks. Meaning, we should check the fault instead of checking all the configuration. That way, in the future, we could combine the queries for all checks with F0467 into one if we want. The reason why the queries are separated for now is to prevent the script from being silent for a long time due to a big query. Even if the total amount of time may be longer, we preferred finishing each check quicker so users can see the progress more often. But we may change this in the future.

With that said, would you mind updating the logic to follow the other checks shown above?

For consistency sake, I would also change the title to `Access (Untagged) Port Config (F0467 native-or-untagged-encap-failure)`.

Also please add pytest and docs. You can see the examples for both from PR #92.

ehaminian:
Hello takishida,

Thanks for the review and the comments.
The requested changes were applied.
Please let me know if there is anything missing.

aci-preupgrade-validation-script.py
```diff
@@ -2824,6 +2824,38 @@ def sup_a_high_memory_check(index, total_checks, tversion, **kwargs):
...
 
 
+def access_untagged_check(index, total_checks, **kwargs):
+    title = 'Access (Untagged) Port Config (F0467 native-or-untagged-encap-failure)'
+    result = PASS
+    msg = ''
+    headers = ["Fault", "POD ID","Node ID","Port","Tenant", "Application Profile", "Application EPG", "Recommended Action"]
+    data = []
+    recommended_action = 'Resolve the conflict by removing this config or other configs using this port in Access(untagged) or native mode.'
+    doc_url = 'https://bst.cloudapps.cisco.com/bugsearch/bug/CSCwj69435'
```
takishida, line 2834:
```suggestion
    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations#native-or-untagged-encap-failure'
```

aci-preupgrade-validation-script.py
```diff
@@ -2824,6 +2824,38 @@ def sup_a_high_memory_check(index, total_checks, tversion, **kwargs):
...
+    
+    faultInsts = icurl('class','faultInst.json?&query-target-filter=wcard(faultInst.changeSet,"native-or-untagged-encap-failure")')
+    fault_dn_regex=r"topology/pod-(?P<podid>\d+)/node-(?P<nodeid>[^/]+)/[^/]+/[^/]+/uni/epp/fv-\[uni/tn-(?P<tenant>[^/]+)/ap-(?P<app_profile>[^/]+)/epg-(?P<epg_name>[^/]+)\]/[^/]+/stpathatt-\[(?P<port>.+)\]/nwissues/fault-F0467"
+    
+    if faultInsts:
+        faultInsts_dn_list = [faultInst['faultInst']['attributes']['dn'] for faultInst in faultInsts]
+        fc = faultInsts[0]['faultInst']['attributes']['code']
+        for faultInst_dn in faultInsts_dn_list:
+            m = re.search(fault_dn_regex, faultInst_dn)
+            if m:
```
takishida, line 2845: Please add `else` to handle a case when the regex is not matching.
Since a fault may be raised against a different type of DN or object, it's common to overlook some corner cases which are not covered by the regex. To avoid those cases being marked as `PASS` incorrectly, we have `unformatted_headers` and `unformatted_data` just to dump the fault description. Please see other F0467 checks for examples.

docs/docs/validations.md
```diff
@@ -66,6 +66,7 @@ Items                                         | Faults         | This Script
 [Different infra VLAN via LLDP][f14]          | F0454: infra-vlan-mismatch | :white_check_mark: | :white_check_mark: 4.2(4) | :white_check_mark:
 [HW Programming Failure][f15]                 | F3544: L3Out Prefixes<br>F3545: Contracts | :white_check_mark: | :white_check_mark: 5.1(1) | :white_check_mark:
 [Scalability (faults related to Capacity Dashboard)][f16] | TCA faults for eqptcapacityEntity | :white_check_mark: | :no_entry_sign: | :white_check_mark:
+[Access (Untagged) Port Config][f17] | F0467: native-or-untagged-encap-failure | :white_check_mark: | :no_entry_sign: | :white_check_mark:
```
takishida, line 69:
```suggestion
[Access (Untagged) Port Config][f17] | F0467: native-or-untagged-encap-failure | :white_check_mark: | :no_entry_sign: | :no_entry_sign:
```

Also would you mind moving this up to right beneath `[L2 Port Config][f8]` as `[Access (Untagged) Port Config][f9]`, then increment the reference number of others (`f9`-`f16`) by 1 to adjust? That way, it'd be easier to comb through the list as all F0467 checks are next to each other.

docs/docs/validations.md
```diff
@@ -83,6 +84,8 @@ Items                                         | Faults         | This Script
 [f14]: #different-infra-vlan-via-lldp
 [f15]: #hw-programming-failure
 [f16]: #scalability-faults-related-to-capacity-dashboard
+[f17]: #native-or-untagged-encap-failure
```
takishida, line 87: Same ask as the above. Move this up as `f9`, please.

docs/docs/validations.md
```diff
@@ -1016,6 +1019,49 @@ Examples of what's monitored via `Operations > Capacity Dashboard > Leaf Capacit
     ```
 
 
+### Native or untagged encap failure
+The APIC GUI or REST previously accepted two different access encapsulations on the same port, despite raising a fault with code F0467 and "native-or-untagged-encap-failure" in the changeSet. This configuration, likely resulting from user error, presents a significant risk of outage during switch upgrades or stateless reloads.
+
+The script verifies these faults to ensure that a port is not configured as part of two access VLANs. It is expected that the customer resolves the conflict causing this fault before any upgrades to prevent potential outages. Failure to do so may result in the deployment of a new VLAN/EPG on the port after the upgrade, leading to downtime in the environment.
+
+!!! example "Fault Example (F0467: native-or-untagged-encap-failure)"
+```
```
takishida, line 1028: This codeblock needs indentations with 4 spaces to be inside the example note.

takishida (changes requested):

docs/docs/validations.md
```diff
@@ -1016,6 +1019,49 @@ Examples of what's monitored via `Operations > Capacity Dashboard > Leaf Capacit
     ```
 
 
+### Native or untagged encap failure
```
takishida, line 1022:
```suggestion
### Access (Untagged) Port Config
```

Let's match the section title with the name in the summary table.

takishida (changes requested):

docs/docs/validations.md
ehaminian: Done!

ehaminian (commented):
codeblock needs indentations with 4 spaces to be inside the example note.

aci-preupgrade-validation-script.py
ehaminian: Done!

ehaminian:
> Same ask as the above. Move this up as `f9`, please.

Also let's use the same title as the table, meaning `#access-untagged-port-config`

ehaminian:
> > Same ask as the above. Move this up as `f9`, please.

> Also let's use the same title as the table, meaning `#access-untagged-port-config` 

> Done!

docs/docs/validations.md
ehaminian: Done!

docs/docs/validations.md
ehaminian: Done!

docs/docs/validations.md
```diff
@@ -58,14 +58,15 @@ Items                                         | Faults         | This Script
...
-[BD Subnets][f10]                             | F0469: duplicate-subnets-within-ctx | :white_check_mark: | :white_check_mark: 5.2(4d) | :white_check_mark:
-[BD Subnets][f11]                             | F1425: subnet-overlap | :white_check_mark: | :white_check_mark: 5.2(4d) | :white_check_mark:
-[VMM Domain Controller Status][f12]           | F0130         | :white_check_mark: | :white_check_mark: 4.2(1) | :white_check_mark:
-[VMM Domain LLDP/CDP Adjacency Status][f13]   | F606391       | :white_check_mark: | :white_check_mark: 4.2(1) | :white_check_mark:
-[Different infra VLAN via LLDP][f14]          | F0454: infra-vlan-mismatch | :white_check_mark: | :white_check_mark: 4.2(4) | :white_check_mark:
-[HW Programming Failure][f15]                 | F3544: L3Out Prefixes<br>F3545: Contracts | :white_check_mark: | :white_check_mark: 5.1(1) | :white_check_mark:
-[Scalability (faults related to Capacity Dashboard)][f16] | TCA faults for eqptcapacityEntity | :white_check_mark: | :no_entry_sign: | :white_check_mark:
+[Access (Untagged) Port Config][9] | F0467: native-or-untagged-encap-failure | :white_check_mark: | :no_entry_sign: | :no_entry_sign:
```
takishida, line 61:
```suggestion
[Access (Untagged) Port Config][f9]            | F0467: native-or-untagged-encap-failure | :white_check_mark: | :no_entry_sign: | :no_entry_sign:
```

docs/docs/validations.md
```diff
@@ -75,14 +76,17 @@ Items                                         | Faults         | This Script
 [f6]: #config-on-apic-connected-port
 [f7]: #l2l3-port-config
 [f8]: #l2l3-port-config
-[f9]: #l3out-subnets
-[f10]: #bd-subnets
+[f9]: #native-or-untagged-encap-failure
```
takishida, line 79:
```suggestion
[f9]: #access-untagged-port-config
```

docs/docs/validations.md
```diff
@@ -664,6 +668,50 @@ It is critical that you resolve these issues before the upgrade to prevent any i
     ```
 
 
+### Access (Untagged) Port Config
+The APIC GUI or REST previously accepted two different access encapsulations on the same port, despite raising a fault with code F0467 and "native-or-untagged-encap-failure" in the changeSet. This configuration, likely resulting from user error, presents a significant risk of outage during switch upgrades or stateless reloads.
+
+The script verifies these faults to ensure that a port is not configured as part of two access VLANs. It is expected that the customer resolves the conflict causing this fault before any upgrades to prevent potential outages. Failure to do so may result in the deployment of a new VLAN/EPG on the port after the upgrade, leading to downtime in the environment.
```
takishida, line 674:
```suggestion
The script verifies these faults to ensure that a port is not configured as part of two access VLANs. You need to resolve the conflict causing this fault before any upgrades to prevent potential outages. Failure to do so may result in the deployment of a new VLAN/EPG on the port after the upgrade, leading to downtime in the environment.
```

takishida (changes requested):

docs/docs/validations.md
ehaminian: Done!

docs/docs/validations.md
ehaminian: Done!

docs/docs/validations.md
ehaminian: Done!

ehaminian: Done!

Issue #74, opened 2023-09-17
New Validation Request: whether many interface override policies configured before upgrading to 5.2(4d) or above


welkin-he:
(use upvote :thumbsup: for attentions)
### Validation Type
[ ] - Fault

[X] - Config

[ ] - Bug

[ ] - Other

### What needs to be validated
For a customer who implemented interface configuration via clis, they may have created many interface override policies. 

### Why it needs to be validated
If customer upgrades the fabric from any version prior to 5.2(4d) to 5.2(4d) and above, this type of configuration can cause APIC shard-20 to take hours before completing postUpgradeCb. If switch upgrade was triggered before APIC complete that, those upgraded switches will experience bootstrap delay and unable to fetch required access policies until APIC completed that. 

This validation is designed to identify if the fabric has similar configuration, so that customer can be alerted.

### Additional context
Add any other context about the feature request here.

takishida:
We need some way to identify whether the postUpgradeCb on APIC is still in progress prior to upgrade switches.

monrog2:
Handled with PostUpgradeCb check. Closing this one.

PR #122, opened 2024-05-15
#74 Check if many interface override policies configured before upgrading to 5.2(4d) or above

monrog2:


monrog2:
Issue being checked for is handled in PostUPgradeCb check. closing this one

PR #148, opened 2024-08-12
CACHECK files upgrade prefix + initial cleanup

monrog2:
Add obvious prefix to files created for APICCA check.
perform initial check and cleanup to prevent existing old files from breaking subsequent runs.


monrog2:
pytest did not change. integration test shows PASS for relevant check

Issue #157, opened 2024-08-27
Check 1, SSH login to APIC fails


dirkf98:
(use upvote :thumbsup: for attentions)
**Describe the bug**
When running the script the first check, "APIC Target version image and MD5 hash", fails.
Based on the logs it looks like the SSH login to the APICs, "ssh user@<apic_ip> -p 22", is successful but the connection fails because the prompt is misinterpreted as `enable` prompt. The script then sends the enable password and is stuck.
Manual ssh login with the same command from CLI works fine.
I'm running the script on an APIC M4 with software version 6.0(2c).

**Script output**
A snippet of the script output that shows the behavior that appears to be a bug

Interestingly, after maybe 10 failed attempts there was one run where it worked on the second APIC, but no the others.
This is the script output:
```
[Check  1/67] APIC Target version image and MD5 hash... 
              Checking apic1......                                                                               ERROR !!
              Checking apic2......                                                                                   DONE
              Checking apic3......                                                                               ERROR !!
              Checking apic4......                                                                               ERROR !!
                                                                                                                  FAIL - UPGRADE FAILURE!!
  APIC                    Firmware  md5sum  Failure                                                   Recommended Action
  ----                    --------  ------  -------                                                   ------------------
  apic3  -         -       ls command via ssh failed due to:failed to login to host  -
  apic4  -         -       ls command via ssh failed due to:failed to login to host  -
  apci1  -         -       ls command via ssh failed due to:failed to login to host  -

```

This is an excerpt from the log. I removed a lot of special characters though, that were not displayed correctly:

```
[2024-08-27 17:57:21.210+0200 DEBUG                __expect:229 ] expect matched result[4] = password
[2024-08-27 17:57:21.210+0200 DEBUG                   login:283 ] matched password prompt, send password
[2024-08-27 17:57:21.261+0200 DEBUG               start_log:158 ] setting logfile to preupgrade_validator_logs/preupgrade_validator_debug.log

Last login: 2024-08-27T17:41:03.000+02:00 UTC
...
!!! Hostname in prompt shows up, but surrounded by a lot of special characters. Copy and paste here does not work
...
[2024-08-27 17:57:38.276+0200 DEBUG                __expect:229 ] expect matched result[8] = timeout
[2024-08-27 17:57:38.276+0200 DEBUG                   login:297 ] timeout received but connection still opened, send enter

...
[...]

PR #160, opened 2024-09-09
relax prompt regex and remove enable

monrog2:
Integration tests passed, so this does not break any existing function. 

Waiting to hear back on the 2 reported cases where this was hit to see if this allows that check to proceed successfully.

### DIFF
--- a/aci-preupgrade-validation-script.py
+++ b/aci-preupgrade-validation-script.py
@@ -122,7 +122,7 @@ class Connection(object):
         self.protocol = "ssh"
         self.port = None
         self.timeout = 30
-        self.prompt = "[^#]#[ ]*(\x1b[\x5b-\x5f][\x40-\x7e])*[ ]*$"
+        self.prompt = "#\s.*$" #"[^#]#[ ]*(.*)*[ ]*$"
         self.verify = False
         self.searchwindowsize = 256
         self.force_wait = 0
@@ -258,7 +258,6 @@ class Connection(object):
             "yes/no": "(?i)yes/no",
             "username": "(?i)(user(name)*|login)[ as]*[ \t]*:[ \t]*$",
             "password": "(?i)password[ \t]*:[ \t]*$",
-            "enable": ">[ \t]*$",
             "prompt": self.prompt
         }
 
@@ -281,18 +280,10 @@ class Connection(object):
             elif match == "password":
                 # don't log passwords to the logfile
                 self.stop_log()
-                if last_match == "enable":
-                    # if last match was enable prompt, then send enable password
-                    logging.debug("matched password prompt, send enable password")
-                    self.child.sendline(self.enable_password)
-                else:
-                    logging.debug("matched password prompt, send password")
-                    self.child.sendline(self.password)
+                logging.debug("matched password prompt, send password")
+                self.child.sendline(self.password)
                 # restart logging
                 self.start_log()
-            elif match == "enable":
-                logging.debug("matched enable prompt, send enable")
-                self.child.sendline("enable")
             elif match == "prompt":
                 logging.debug("successful login")
                 self._login = True


takishida (commented):
It's a bit of nitpicking, but let's remove all the other lines for `enable_password` in class `Connection` as well since they are no longer used, including the comments.

What I could find were these:

In the class description comment:
```python
class Connection(object):
    """
    Object built primarily for executing commands on Cisco IOS/NXOS devices.  The following
      --- omit ---
        enable_password  (opt) enable password credential (IOS only) (default 'cisco')```
    """
```

In `__init__`:
```python
      def __init__(self, hostname):
     --- omit ---
          self.enable_password = 'cisco'
```

In the comment of `login()`:
```python
          # check for user provided 'prompt' which indicates successful login
          # else provide approriate username/password/enable_password     <---
          matches = {
```

takishida:
Just for the record, let me share a bit more information on what this PR is trying to solve.

On APIC, the default login shell is not bash. It's a proprietary shell called "loginshell". This is using zsh under the hood.

zsh utilizes many ANSI escape sequences even with minimum settings. Here is an example.

### Normal zsh example
This is what we see on our terminal:
```
ubuntu%
ubuntu%
ubuntu%
```

But with ANSI escape sequences displayed as is, the terminal output above looks like this:
```
\x1b[1m\x1b[7m%\x1b[27m\x1b[1m\x1b[0m    \r \r\r\x1b[0m\x1b[27m\x1b[24m\x1b[Jubuntu% \x1b[K\x1b[?1h\x1b=\x1b[?2004h\x1b[?1l\x1b>\x1b[?2004l
\x1b[1m\x1b[7m%\x1b[27m\x1b[1m\x1b[0m    \r \r\r\x1b[0m\x1b[27m\x1b[24m\x1b[Jubuntu% \x1b[K\x1b[?1h\x1b=\x1b[?2004h\x1b[?1l\x1b>\x1b[?2004l
\x1b[1m\x1b[7m%\x1b[27m\x1b[1m\x1b[0m    \r \r\r\x1b[0m\x1b[27m\x1b[24m\x1b[Jubuntu% \x1b[K\x1b[?1h\x1b=\x1b[?2004h\x1b[?1l\x1b>\x1b[?2004l
```

### APIC shell example

The same is happening on APIC's shell. Here is an example of the first output after you login via SSH.
```
Last login: 2024-09-11T01:49:27.000-04:00 UTC
apic1# 
```

With ANSI escape sequences displayed as is, it looks like this:
```
Last login: 2024-09-11T01:49:27.000-04:00 UTC
\x1b[1m\x1b[7m%\x1b[27m\x1b[1m\x1b[0m     \r \r\r\x1b[0m\x1b[27m\x1b[24m\x1b[Japic1# \x1b[K\x1b[?1h\x1b=\x1b[?2004h\x1b[?1l\x1b>\x1b[?2004l
```

But sometimes, the data sent over to our client, in our case the script, may stop half way until further input is sent from us. And it looks like this.
```
Last login: 2024-09-11T01:49:27.000-04:00 UTC
\x1b[1m\x1b[7m%\x1b[27m\x1b[1m\x1b[0m     \r \r\r\x1b[0m\x1b[27m\x1b[24m\x1b[Japic1# \x1b[K
```

In the latter case, the current prompt regex (`[^#]#[ ]*(\x1b[\x5b-\x5f][\x40-\x7e])*[ ]*$`) works. But in the former case, it does not work.

The log reported in issue #157 contains many null data (`\x00`) but the idea is still the same.

This is why relaxing the prompt regex to simply "# " worked.

We could also trim the ANSI escape sequences by overloading some internal functions of pexpect module, but as we need to support both python 2 and 3 as some old APIC versions use py2, it's not easy to find a solution that works for both.

takishida (approved):

PR #169, opened 2024-10-01
adding option to give the values of user pass and image as parameter

adityaprayas:


monrog2:
Summarizing what @takishida shared:

1. What is the issue this is trying to fix? Please file a matching GIT issue documenting this.
2. update the matching documentation with usage and why someone would supply these parameters instead of relying on what is already present within the firmware repo as the script does today
3. ensure that pytests are written out for the new logic to cover both with/without/any foreseen failure scenarios, such as if user enters a version that does not exist or follow expected formatting.

monrog2:
Have not heard any updates on this one, closing out without merging.

Issue #173, opened 2024-10-10
Overlapping VLAN pool check fails due to "API call failed" when there are too many VLAN deployments


takishida:
(use upvote :thumbsup: for attentions)
**Describe the bug**
The overlapping VLAN pool check fails because the API call failed when there are too many objects to be returned, precisely speaking more than 100K.

The observed failure in the overlapping VLAN pool check was specific to an API query for object `fvIfConn` which represents VLAN deployments on each interface. However, the same failure could happen with any other API queries in any check.

**Script output**
```
[Check 35/67] Overlapping VLAN Pools...
              Error: API call failed! Check debug log...                                                                          ERROR !!
```

**To Reproduce**
Run the script on an ACI fabric with more than 100K instances for an object that is queried in any check of the script.

**Expected behavior**
The script should use pagination to query objects 100K or less at a time instead of trying to query everything at once.

**Additional context**
In the script log (`preupgrade_validator_debug.log`), there is an API call failure reported due to the result dataset being too big.

```
[2024-10-10 16:54:51.504+0000 INFO                    icurl:1053] cmd = icurl -gs http://127.0.0.1:7777/api/class/fvIfConn.json
[2024-10-10 16:54:58.420+0000 DEBUG                   icurl:1055] response: {"totalCount":"1","imdata":[{"error":{"attributes":{"code":"400","text"
:"Unable to process the query, result dataset is too big"}}}]}

[2024-10-10 16:54:58.421+0000 ERROR                <module>:4343] API call failed! Check debug log
Traceback (most recent call last):
  File "aci-preupgrade-validation-script.py", line 4331, in <module>
    r = check(idx + 1, len(checks), **inputs)
  File "aci-preupgrade-validation-script.py", line 2015, in overlapping_vlan_pools_check
    fvIfConns = icurl('class', 'fvIfConn.json')
  File "aci-preupgrade-validation-script.py", line 1065, in icurl
    raise Exception('API call failed! Check debug log')
Exception: API call failed! Check debug log
```

PR #174, opened 2024-10-10
Add pagination in `icurl()` to support large amount of objects

takishida:
The current API query example made by `icurl()`:
```
/api/class/fvIfConn.json
```

The new API query example with pagination:
```
/api/class/fvIfConn.json?page=0&page-size=100000
```

With the pagination, the API query in `icurl()` will query up to 100K object instances at a time. If there are more objects to be returned, it will automatically query again with the next page such as `page=1`, `page=2` and so on, then it will concatenate all the  result as a single list so that the output of `icurl()` remains the same regardless of the pagination or not.

This enhancement will fix issue #173 and any other failure due to massive amount of objects to be returned.

### DIFF
--- a/aci-preupgrade-validation-script.py
+++ b/aci-preupgrade-validation-script.py
@@ -1044,6 +1044,18 @@ def print_result(title, result, msg='',
     prints(output)
 
 
+def _icurl_error_handler(imdata):
+    if imdata and "error" in imdata[0]:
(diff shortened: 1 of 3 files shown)


monrog2 (approved):
Discussed API-only commSetup.masMos lookup, decided to move forward as-is as that change appears to be API-only today.

PR #155, opened 2024-08-17
Enhance Overlapping VLAN pools check

takishida:
The current `overlapping_vlan_pools_check` only checks whether each EPG has multiple domains and whether or not those domains have VLAN pools with overlapping VLAN IDs. It does not check the VLAN or interfaces used in those EPGs.
The current implementation can warn users about the potential risk of overlapping VLAN pools and highly encourage them to check their design and implementation. However, this can be a daunting task.

The enhancement in this PR checks the VLAN and interfaces in those EPGs along with the VLAN scope (portlocal/global) to show users the exact VLAN ID and interfaces that are using the overlapping VLAN pools along with their impact. This can reduce the amount of configurations to be examined by users, and can also prevent a potential false alarm that may occur with the current implementation, such as when the EPG does have overlapping VLAN pools but the VLAN used in the EPG is not in the overlapping range.

In the examples below, `TN1:AP1:EPG1` is listed as one of the problematic ones in the current implementation. However, it is not listed in the new output with the enhancement because the VLAN used in the EPG is not in the overlapping ranges of those pools.
The new output also shows different impacts for each VLAN and interface. The main difference is whether the affected VLAN is implemented on a vPC or not. In such as case, the impact is shown as `Outage` because it affects not only the flooding scope but also the endpoint learning. More detailed explanations and examples were added in the [github doc](https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#overlapping-vlan-pool) with this PR.

**Current Output Example**
```
[Check 35/67] Overlapping VLAN Pools...                                                                            FAIL - OUTAGE WARNING!!
  Tenant  AP   EPG   VLAN Pool (Domain) 1  VLAN Pool (Domain) 2  Recommended Action
  ------  --   ---   --------------------  --------------------  ------------------
  TN1     AP1  EPG1  VLAN_Pool1 (PhyDom1)  VLAN_Pool2 (PhyDom2)  Resolve overlapping VLANs between these two VLAN pools
  TN2     AP1  EPG1  VLAN_Pool3 (PhyDom3)  VLAN_Pool4 (PhyDom4)  Resolve overlapping VLANs between these two VLAN pools
  TN3     AP1  EPG3  VLAN_Pool5 (PhyDom5)  VLAN_Pool6 (PhyDom6)  Resolve overlapping VLANs between these two VLAN pools

  Reference Document: "Overlapping VLAN Pool" from from Pre-Upgrade Check Lists
```

**New Output Example**
```
[...]

PR #182, opened 2024-11-07
Remove vnid_mismatch_check

takishida:
`vnid_mismatch_check` was added to complement `overlapping_vlan_pools_check`.

Given the recent update (#155, #162), `vnid_mismatch_check` became redundant, and it has a risk of raising a false alarm (see below for details). Thus, it was decided to be removed.

An example of false alarm from `vnid_mismatch_check`:

```
Node 101 --- EPG A (VLAN 10) --- domain A --- VLAN pool A (VLAN 10)
Node 105 --- EPG A (VLAN 10) --- domain B --- VLAN Pool B (VLAN 10)
```

Note that these two switches are not vPC pair.

This is alerted by `vnid_mismatch_check` because VLAN 10 on node 101 and 105 are assigned a different VNID from respective VLAN pools. However, it's expected and will not cause any issues with an upgrade because the VNID for each node is deterministic and will always be the same even after the upgrade. See the "Example:  good only for a specific use case" in https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#overlapping-vlan-pool for details.

### DIFF
--- a/aci-preupgrade-validation-script.py
+++ b/aci-preupgrade-validation-script.py
@@ -2176,68 +2176,6 @@ def overlapping_vlan_pools_check(index, total_checks, **kwargs):
     return result
 
 
-def vnid_mismatch_check(index, total_checks, **kwargs):
-    title = 'VNID Mismatch'
-    result = FAIL_O
-    msg = ''
-    headers = ["EPG", "Access Encap", "Node ID", "Fabric Encap"]
-    data = []
-    mismatch_hits = []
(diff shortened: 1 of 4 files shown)


monrog2 (commented):
Code changes look good, lets discuss and finalize how we want to handle documentation around checks we remove completely.

takishida:
Doc update preview

image

image

monrog2 (approved):
doc changes also look good. merging

PR #184, opened 2024-12-01
fixed openssl path issue.

welkin-he:
removed /bin/ path for openssl and tested in 5.2(8i) which is fine.


monrog2 (approved):
small enough change, quick test on lab fabs show that openssl cmd took without issue. merging

PR #205, opened 2025-03-25
check by dn, not name. also ignore inactive nodes

monrog2:


monrog2:
local tests look better after this one. merging.

PR #208, opened 2025-03-25
truncate stdoutput to just action items

monrog2:


monrog2:
no pytests were harmed in the making of this PR:
`426 passed, 1 warning in 0.43s `

monrog2:
An alternative is something along the lines of:
```python
def prints(objects, sep=' ', end='\n', result="pending"):
    with open(RESULT_FILE, 'a') as f:
        if result in [FAIL_O, FAIL_UF, MANUAL, ERROR, POST]:
            print(objects, sep=sep, end=end, file=sys.stdout)
            sys.stdout.flush()
        if result != "pending":
            print(objects, sep=sep, end=end, file=f)
            f.flush()

def print_title(title, index=None, total=None, result="pending"):
    if index and total:
        end = '\r' if result== "pending" else ''
        prints('[Check{:3}/{}] {}... '.format(index, total, title), end=end, result=result)
    else:
        prints('{:14}{}... '.format('', title), end='')


def print_result(title, result, msg='',
                 headers=None, data=None,
                 unformatted_headers=None, unformatted_data=None,
                 recommended_action='',
                 doc_url='',
                 adjust_title=False, index=None, total_checks=None):
    padding = 120 - len(title) - len(msg)
    if adjust_title: padding += len(title) + 18
    output = '{}{:>{}}'.format(msg, result, padding)
    if data:
        data.sort()
        output += '\n' + format_table(headers, data)
    if unformatted_data:
        unformatted_data.sort()
        output += '\n' + format_table(unformatted_headers, unformatted_data)
    if data or unformatted_data:
        output += '\n'
        if recommended_action:
            output += '\n  Recommended Action: %s' % recommended_action
        if doc_url:
            output += '\n  Reference Document: %s' % doc_url
        output += '\n' * 2
    if index and total_checks and result in [FAIL_O, FAIL_UF, MANUAL, ERROR, POST]:
        print_title(title, index, total_checks, result=result)
    prints(output, result=result)
```
and then adding the index/total checks to each print_results treating it as the final write for each check to both stdout is an error, and the file for logging.

In both cases, the connections check re-using these functions makes it messy.

monrog2:
closing, not putting cycles on this with PUV on the way.

PR #210, opened 2025-03-26
195 attribute field bitinfo missing from json in 32 and 64 bit firmware image for switches

monrog2:


monrog2:
fixed:
1. bitinfo not available on pre 6.0(2) versions > cversion less_than 6.0(2) will return POST
2. cloudsec check failing due to old version > try/catch the failure for older versions
3. EEDCH check returning blank for pre-5.2 versions > this is expected regardless of cipher changes. Thread indicates that main issue occurs on 5.2, but the relality is that MOs don't exist for commCiphers until after upgrade, a bit of chicken/egg

monrog2:
lab tests are now cleared of errors

monrog2:
all pytests pass

PR #247, opened 2025-06-03
Add ISIS DTEP Check

joelebla:
Add ISIS DTEP Check for CSCwp15375


monrog2 (approved):
integration test passed, merging

PR #276, opened 2025-08-12
Validate Preferred Group shared-service providers (#244)

### LINKED ISSUE #244: NewValidation: Shared Services Provider and Preferred Group not supported post-upgrade
### Validation Type
[ ] - Fault

[ x ] - Config

[ ] - Bug

[ ] - Other

1) Upgrading to 4.2(6d) and later or 5.1(1h) and later
2) EPG is a shared services provider (global pctag, provided global contract) and is part of a preferred group

Due to the behavior change contracts will no longer be pushed post-upgrade causing a significant datacenter outage.

The configuration will be faulted: e.g. F0467 Configuration failed for uni/tn-common/brc-shared/contract/epgCont-1/instp-ext-epg due to Invalid Contract Configuration, debug message: invalid-contract-config: Shared service provider cannot be in a Preferred Group.

### Additional context
[Documentation](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/aci/apic/sw/5-x/aci-fundamentals/cisco-aci-fundamentals-50x/m_policy-model.html)

"Due to CSCvm63145, an EPG in a Contract Preferred Group can consume a shared service contract, but cannot be a provider for a shared service contract with an L3Out EPG as consumer."

The configuration will be faulted starting in 4.2(6d)+ and 5.1(1h)+.

jeestr4d:
Fixes #244.

## Summary

Adds a pre-upgrade validation for Preferred Group members that provide tenant-
or global-scope shared-service contracts.

The check:

- Runs for ACI target releases 4.2 and later.
- Selects `fvAEPg` and `l3extInstP` providers with
  `prefGrMemb="include"` and shared/global pcTags `17-16385`.
- Requires a materialized provider-side `vzFromEPg` to `vzToEPg`
  relationship; provider configuration alone is not reported.
- Resolves provider context through
  `vzFromEPg.scopeId -> fvCtxDef.scope -> fvCtxDef.dn`.
- Compares the provider context DN with `vzToEPg.ctxDefDn` and reports only
  cross-context relationships.
- Requires matching tenant ancestry for tenant-scope contracts.
- Reports the specific affected consumer DN and de-duplicates result rows.

## Version behavior

| Target version | Behavior |
|---|---|
| Before 4.2 | Check is not applicable |
| 4.2 through target releases before 6.0(1g) | Report any materialized cross-context consumer class; early releases can have silent forwarding risk, while later 4.2/5.x releases can reject the configuration with F0467 |
| 6.0(1g) and later | Report only cross-context L3Out (`/instP-`) or `vzAny` (`/any`) consumers; ordinary EPG-to-EPG shared service is allowed |
| 6.1(3f) and later | Same restricted topology; product validation may surface the condition through F4684 |

The source-proven Preferred Group-specific F0467 boundaries are 4.2(6d),
5.1(3e), and first-proven 5.2(1g). The first source-visible narrowed behavior
is 6.0(1g).

## Sanitized integration result

The deterministic ACI 5.2(4d) to 5.2(8f) integration job identified seven
materialized cross-context relationships across three fabrics:

| Fabric | Findings |
|---|---:|
| Fabric A | 2 |
| Fabric B | 4 |
| Fabric C | 1 |

Sanitized examples:

| Shared-service contract | Preferred Group provider | pcTag | Affected consumer |
|---|---|---:|---|
| `uni/tn-Tenant-A/brc-Shared-Contract-1` | `uni/tn-Tenant-A/ap-App-Provider/epg-Provider-1` | 57 | `uni/tn-Tenant-B/ap-App-Consumer/epg-Consumer-1` |
| `uni/tn-Tenant-C/brc-Shared-Contract-2` | `uni/tn-Tenant-C/ap-App-Provider/epg-Provider-2` | 10934 | `uni/tn-Tenant-D/ap-App-Consumer/epg-Consumer-2` |
| `uni/tn-common/brc-Shared-Contract-3` | `uni/tn-Tenant-E/ap-App-Provider/epg-Provider-3` | 10932 | `uni/tn-Tenant-F/ap-App-Consumer/epg-Consumer-3` |

Every reported row has a concrete materialized consumer instead of the prior
provider-only `Any` value.

One additional integration fabric returned APIC DME `503 Resolve timeout`
responses. Multiple unrelated validations timed out concurrently on that
fabric. In most affected runs, this check timed out on its first `vzBrCP`
query and never reached the new `fvCtxDef` or `vzFromEPg` queries.

## Validation

- Focused Preferred Group tests: 35 passed.
- Full local suite: 1,050 passed.
- GitLab Python 2.7 unit job: passed.
- GitLab Python 3.8 unit job: passed.
- GitLab integration pipeline exercised the updated logic against live
  fabrics.


monrog2:
integration test has no issues related to this check

aci-preupgrade-validation-script.py
```diff
@@ -6056,6 +6056,62 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
...
+    data = []
+    recommended_action = 'an EPG in a Contract Preferred Group can consume a shared service contract, but cannot be a provider for a shared service contract.'
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#preferred_group_shared_service_provider'
+
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+    # Configuration becomes faulted after 4.2-6d and 5.1-1h
+    if tversion.older_than("4.2(6d)"):
+        return Result(result=NA)
+    elif tversion.older_than("5.1(1h)"):
```
monrog2, line 6072: [P1] Gate these thresholds by release train. With the current sequential comparisons, every 4.2 target from 4.2(6d) onward passes the first check but is still older than 5.1(1h), so this returns N/A for an affected upgrade. I reproduced that behavior with target 4.2(6d). Please model the 4.2(6d)+ and 5.1(1h)+ branches explicitly and add boundary tests for affected 4.2 releases.

aci-preupgrade-validation-script.py
```diff
@@ -6056,6 +6056,62 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
...
+        list_of_shrd_contracts.append(shrd_contract["vzBrCP"]["attributes"]["dn"])
+    # Configuration only becomes faulted for extEPGs after 6.0(1g), normal epgs permitted.
+    if tversion.older_than("6.0(1g)"):  
+        glbl_epgs_api = 'fvAEPg.json'
+        glbl_epgs_api += '?query-target-filter=and(le(fvAEPg.pcTag,"16385"),ge(fvAEPg.pcTag,"16"),eq(fvAEPg.prefGrMemb,"include"))'
+        glbl_epgs_api += '&rsp-subtree=children&rsp-subtree-class=fvRsProv'
+        glbl_epgs = icurl('class', glbl_epgs_api)
+        if glbl_epgs:
+            for glbl_epg in glbl_epgs:
+                for prov_contract in glbl_epg["fvAEPg"]["children"]:
```
monrog2, line 6090: [P1] Treat `children` as optional here and in the l3extInstP loop below. APIC omits the key when a preferred-group object has no provider relations; direct indexing then raises `KeyError`, converts this check to ERROR, and can prevent later affected objects from being evaluated. I reproduced `ERROR !! Unexpected Error: children` with a valid childless EPG response. Iterating over `.get("children", []) or []` avoids that failure, and a mixed childless/affected fixture should cover it.

aci-preupgrade-validation-script.py
```diff
@@ -6056,6 +6056,62 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
...
+                    if prov_contract["fvRsProv"]["attributes"]["tDn"] in list_of_shrd_contracts:
```
monrog2, line 6091: [P1] Correlating only the provider with a global-scope contract is broader than the documented defect condition. CSCvm63145 and the new documentation require an L3Out EPG consumer in another VRF, but this condition flags any preferred-group provider of a global contract, including contracts without an L3Out consumer. Please query/correlate the consumer relation before reporting FAIL_O and add PASS cases for no consumer or a non-L3Out consumer.

aci-preupgrade-validation-script.py
```diff
@@ -6056,6 +6056,62 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
     return Result(result=result, headers=headers, data=data, recommended_action=recommended_action, doc_url=doc_url)
 
 
+@check_wrapper(check_title='Shared Services Providers with Preferred Group enabled')
+def pg_and_shared_svc_contract_check(cversion, tversion, **kwargs):
+    result= PASS
+    headers = ["Shared Service Contract", "Provider in Preferred Group", "PcTag"]
+    data = []
+    recommended_action = 'an EPG in a Contract Preferred Group can consume a shared service contract, but cannot be a provider for a shared service contract.'
```
monrog2, line 6064: [P2] This recommendation restates the restriction but does not tell an operator how to make the upgrade safe. Please provide an actionable remediation, such as removing the EPG from the preferred group or stopping it from providing the shared-service contract, and verify the recommendation in the tests.

tests/checks/pg_and_shared_svc_contract_check/test_pg_and_shared_svc_contract_check.py
```diff
@@ -0,0 +1,146 @@
...
+            script.PASS,
+        ),
+    ]
+)
+def test_logic(run_check, mock_icurl, cversion, tversion, expected_result):
+    result = run_check(
+        cversion=script.AciVersion(cversion),
+        tversion=script.AciVersion(tversion) if tversion else None
+    )
+    assert result.result == expected_result
```
monrog2, line 146: [P2] The fixture directory and parametrized structure match the repository convention, but asserting only `result.result` allowed the release-gate, missing-children, and consumer-correlation gaps to pass. Please add focused cases for affected 4.2 boundaries, childless API objects, and contracts with/without L3Out consumers, and assert the relevant `data` and recommendation fields.

monrog2 (changes requested):
Requesting changes because the new validation can skip affected upgrade targets, fail on valid APIC responses, and report outage warnings for configurations outside the documented defect condition. The test layout follows the repository convention, but it needs regression coverage for these cases and should validate the returned details, not only the status. Focused Python 3.8 tests pass (10/10), but the version-gating and missing-children defects were independently reproduced.

PR #293, opened 2025-10-31
NewValidation for CSCwp91797

### LINKED ISSUE #285: NewValidation: CSCwp91797
### Validation Type
[ ] - Fault

[ ] - Config

[X] - Bug

[ ] - Other

CSCwp91797 - the condition can be found from the bug.

all vpc can went down if customer's is not aware of this issue.

### Additional context
Add any other context about the feature request here.

welkinhe:
Addresses #285.

This PR remains open for rebase and correction of the affected-version logic against the current CSCwp91797 defect record.

Fixes #285


monrog2 (changes requested):
Details on changes below

aci-preupgrade-validation-script.py
```diff
@@ -5558,6 +5558,33 @@ def configpush_shard_check(tversion, **kwargs):
...
+
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING) 
+
+    if tversion.same_as("6.0(9d)"):
+        result = PASS
+        port_tracking_api = 'uni/infra/trackEqptFabP-default.json'
+        port_tracking_mo = icurl('mo', port_tracking_api)
+        if port_tracking_mo:
+            admin_st = port_tracking_mo[0]['infraPortTrackPol']['attributes']['adminSt']
```
monrog2, line 5577: **[Blocking] Empty policy response produces a false PASS**

`result` is set to `PASS` before this query, so an empty successful response leaves the check passing even though the required singleton policy was never evaluated. The same path also assumes exactly one correctly shaped object. Treat empty, duplicate, wrong-class, or attribute-incomplete responses as `ERROR`; only return `PASS` after evaluating exactly one valid `infraPortTrackPol`. Add empty, malformed, and duplicate-response tests that assert both `Result` and `AciResult` output.

aci-preupgrade-validation-script.py
```diff
@@ -5558,6 +5558,33 @@ def configpush_shard_check(tversion, **kwargs):
...
+    result = NA
+    headers = ["Admin State", "Port Tracking Active Fabric Ports"]
+    data = []
+    recommended_action = 'Increase Port Tracking Active Fabric Ports to 1 before upgrade'
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#port-tracking-active-fabric-port-zero'
+
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING) 
+
+    if tversion.same_as("6.0(9d)"):
```
monrog2, line 5572: **[Blocking] Confirmed affected releases are classified N/A**

This exact-version gate skips `6.1(3f)`, which the authoritative CSCwp91797 release note explicitly identifies as affected, and the record indicates additional affected releases exist. Those targets currently return `N/A` without evaluating the policy, hiding an outage risk. Reconcile the train-specific affected/fixed matrix with the defect owner, encode all authoritative boundaries, update the documentation, and test `6.0(9d)`, `6.1(3f)`, each known fixed boundary, and adjacent releases.

aci-preupgrade-validation-script.py
```diff
@@ -5558,6 +5558,33 @@ def configpush_shard_check(tversion, **kwargs):
 
     return Result(result=result, headers=headers, data=data, recommended_action=recommended_action, doc_url=doc_url)
 
+@check_wrapper(check_title='Port Tracking Minimal Uplink Zero')
+def port_tracking_active_fabric_port_check(tversion, **kwargs):
+    result = NA
+    headers = ["Admin State", "Port Tracking Active Fabric Ports"]
+    data = []
+    recommended_action = 'Increase Port Tracking Active Fabric Ports to 1 before upgrade'
```
monrog2, line 5566: **[High] Recommended change omits its safety prerequisite**

The authoritative CSCwp91797 guidance makes `minlinks=1` conditional on sufficient fabric-uplink redundancy and provides disabling port tracking before upgrade as an alternative. This unconditional instruction can cause service impact on an unsuitable leaf. Include the prerequisite and supported alternatives, or direct users to the authoritative procedure/TAC when suitability cannot be verified. Assert the complete recommendation and result payload in tests.

aci-preupgrade-validation-script.py
```diff
@@ -5558,6 +5558,33 @@ def configpush_shard_check(tversion, **kwargs):
 
     return Result(result=result, headers=headers, data=data, recommended_action=recommended_action, doc_url=doc_url)
 
+@check_wrapper(check_title='Port Tracking Minimal Uplink Zero')
+def port_tracking_active_fabric_port_check(tversion, **kwargs):
```
monrog2, line 5562: **[Medium] No-vPC fabrics receive a false outage warning**

CSCwp91797 affects vPC physical member ports, but this check does not consume the already available `vpc_node_ids`. A fabric with no vPC can therefore receive `FAIL_O` and unnecessary remediation guidance. Accept `vpc_node_ids`, return `N/A` before querying when no vPC topology exists, and add tests for empty and populated vPC inventories.

monrog2 (changes requested):
Please address the four inline findings before merge.

PR #307, opened 2025-12-22
added validation for CFD CSCwp64296

psureshb:


diwakarmishra1190 (approved):
LGTM 
tested Locally

diwakarmishra1190 (approved):
New changes are working fine tested locally.

aci-preupgrade-validation-script.py
```diff
@@ -6012,6 +6012,48 @@ def apic_vmm_inventory_sync_faults_check(**kwargs):
...
+    # Target version check
+    if not tversion:
+        prints("Target version not provided, skipping check.")
+        return Result(result=MANUAL, msg=TVER_MISSING)
+    
+    # Affected source version is in range [5.2(3):6.0(3)] . Fixed on 6.0(9e)+ and 6.1(4)+.
+    # if cversion.newer_than("3.1(2v)") and tversion.older_than("6.1(3g)"):
+    if (
+    (cversion.same_as("5.2(3e)") or cversion.newer_than("5.2(3e)")) and
+    (cversion.same_as("6.0(3g)") or cversion.older_than("6.0(3g)")) and
```
takishida: Why 6.0(3g)? The enhancement that caused CSCwp64296 was introduced in the first release of 6.0(3) which is 6.0(3d).

aci-preupgrade-validation-script.py
```diff
@@ -6012,6 +6012,48 @@ def apic_vmm_inventory_sync_faults_check(**kwargs):
...
+    (
+        tversion.older_than("6.0(9e)") or
+        ((tversion.same_as("6.1(1f)") or tversion.newer_than("6.1(1f)")) and tversion.older_than("6.1(4h)"))
+    )
+    ):
+        # endpoint to fetch the rogue exception MACs
+        exception_mac_api = 'fvRogueExceptionMac.json?query-target-filter=and(wcard(fvRogueExceptionMac.dn,"([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2}"))'
```
takishida: Why are you matching MAC address in regex here? `fvRogueExceptionMac` is for MAC as the name suggests. Its DN always has a MAC address.

Unless I'm missing something, this is just causing unnecessary query work on the APICs.

aci-preupgrade-validation-script.py
```diff
@@ -6012,6 +6012,48 @@ def apic_vmm_inventory_sync_faults_check(**kwargs):
...
+
+        # endpoint to fetch the presListener entries
+        presListener_api = 'presListener.json?query-target-filter=and(wcard(presListener.dn,"exceptcont"))'
```
takishida: Instead of regex(`wcard`)), you should do exact match with `lstDn` such as `eq(presListener.lstDn,"exceptcont")`

And use `rsp-subtree-include=count` instead of having APICs return all of the contents because what we need to know is only the number of objects.

aci-preupgrade-validation-script.py
```diff
@@ -6012,6 +6012,48 @@ def apic_vmm_inventory_sync_faults_check(**kwargs):
...
+
+        exception_macs = icurl('class', exception_mac_api)
+
+        if exception_macs:
+            prints("Found {} exception MACs, checking presListener entries...".format(len(exception_macs)))
```
takishida: Do not use `prints()` inside a check function.

In the new framework with the progress bar, it will end up like this, breaking the progress bar view. Please use logging instead, if needed. Otherwise, those info should be stored in the `data` so that it can be displayed in the result table and JSON file.

```
--- skip ---
Collecting VPC Node IDs...

Found 1 exception MACs, checking presListener entries...-------------------------------------------------------| 0/1 checks completed     <--- !!!!!
Progress: |████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 checks completed


=== Check Result (failed only) ===
```

aci-preupgrade-validation-script.py
```diff
@@ -6012,6 +6012,48 @@ def apic_vmm_inventory_sync_faults_check(**kwargs):
...
+def rogue_ep_coop_exception_mac_check(cversion, tversion, **kwargs):
+    result = PASS
+    headers = ["Rogue Exception MACs Count", "presListener Count"]
+    data = []
+    recommended_action = 'Remove the affected EP exception configurations and re-add them'
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#roguecoop-exception-list-missing-on-switches'
+
+    # Target version check
+    if not tversion:
+        prints("Target version not provided, skipping check.")
```
takishida:
```suggestion
```

takishida (changes requested):
@psureshb  @diwakarmishra1190 

I've updated the doc with the information that is relevant to users along with the new conditions.

Correct me if I'm wrong, but I doubt that the current condition checked by this rule will ever hit because I believe `presListener` will be missing on APICs after APICs are upgraded. But you are checking the current version being the older one and the target version being the affected one. That is only the case BEFORE APICs (or anything) are upgraded. In other words, the `presListener` issue has yet to happen.

Or if `presLIstener`s are created on APICs when switches are upgraded instead of APICs, using that as a condition doesn't make sense at all. This script is to prevent the issue from happening by detecting the conditions before the issue happens instead of checking the symptoms of the issue after hitting it.

Can you please review the updated doc and update the rule logic accordingly?

aci-preupgrade-validation-script.py
psureshb: version changed to 6.0(3d)

aci-preupgrade-validation-script.py
psureshb: removed the logic of matching MAC address using regex

aci-preupgrade-validation-script.py
psureshb: using `eq(presListener.lstDn,"exceptcont")`  and `rsp-subtree-include=count`

aci-preupgrade-validation-script.py
psureshb: removed prints staments

psureshb:
[pytest logs.txt](https://github.com/user-attachments/files/25971211/pytest.logs.txt)
[preupgrade_validator_logs.zip](https://github.com/user-attachments/files/25971306/preupgrade_validator_logs.zip)

docs/docs/validations.md
```diff
@@ -2668,6 +2672,33 @@ To avoid this risk, consider disabling Auto Firmware Update before upgrading to
     This issue occurs because older switch firmware versions are not compatible with switch images 6.0(3) or newer. The APIC version is not a factor.
 
 
+### Rogue EP Exception List missing on switches
+
+Rogue Endpoint Control and COOP Dampening are features that mitigate the impact of flapping endpoints by temporarily pausing the learning of such endpoints. However, in some environments, certain MAC or IP addresses are expected to move frequently.
```
lovkeshsharma702: remove 2677, no need to educate Cu on basic stuff. also remove all extra stuff.  just keep info related to bug. 

"The **Rogue/COOP Exception List** feature, introduced in 5.2(3), allows you to exclude specific MAC addresses from Rogue Endpoint Control and COOP Dampening. Initially, each MAC address had to be configured individually in each bridge domain. In 6.0(3), this feature was enhanced to support fabric-wide exception lists with wildcard options per bridge domain and the ability to exclude MAC addresses in L3Outs. 

due to [CSCwp64296][64], when upgrading spine switches to version 6.0(3)+ from an older version with **Rogue/COOP Exception List**s configured, some exception lists config may not be pushed to the spine switches after reboot. As a result, the feature may stop functioning after the upgrade. 

    The root cause is that internal objects called `presListener` for **Rogue/COOP Exception List**, which publish the configuration from APICs to switches, may be missing on the APICs after an upgrade. This is due to [CSCwp64296][64], introduced with the enhancement in 6.0(3).

Recommended action: delete the affected exception list and create it again. If needed, contact Cisco TAC to help recover missing `presListener` objects on APICs. 
" Keep only relevent info. No extra symbol or AI marking.

tests/checks/rogue_ep_coop_exception_mac_check/test_rogue_ep_coop_exception_mac_check.py
```diff
@@ -0,0 +1,212 @@
...
+        (
+            {exception_mac_api: read_data(dir, "rogue_mac_response.json"),
+             presListener_api: read_data(dir, "presListener_exceptcont.json")},
+            "6.1(2f)",
+            "6.1(2f)",
+            script.PASS,
+        ),
+    ],
+    ids=[
+        "PASS_non_affected_tversion_6.1(4h)_fixed_affected_cversion_no_exception_MACs",
```
lovkeshsharma702: follow the structure used in old bug.

lovkeshsharma702 (commented):
work on comments.

tests/checks/rogue_ep_coop_exception_mac_check/test_rogue_ep_coop_exception_mac_check.py
psureshb: removed the ids for compliance with the standard structure used in other pytest files.

docs/docs/validations.md
psureshb: removed line 2677
Removed the additional data.

docs/docs/validations.md
```diff
@@ -2668,6 +2672,17 @@ To avoid this risk, consider disabling Auto Firmware Update before upgrading to
     This issue occurs because older switch firmware versions are not compatible with switch images 6.0(3) or newer. The APIC version is not a factor.
 
 
+### Rogue EP Exception List missing on switches
+
+The Rogue/COOP Exception List feature, introduced in 5.2(3), allows exclusion of specific MAC addresses from Rogue Endpoint Control and COOP Dampening. Initially, each MAC address had to be configured individually in each bridge domain. In 6.0(3), this feature was enhanced to support fabric-wide exception lists with wildcard options per bridge domain and the ability to exclude MAC addresses in L3Outs.
+
+However, due to [CSCwp64296][64], when upgrading spine switches to version 6.0(3)+ from an older version with Rogue/COOP Exception Lists configured, some exception lists may not be pushed to the spine switches. As a result, the feature may stop functioning after the upgrade. 
+
+The root cause is that internal objects called `presListener` for Rogue/COOP Exception List, which publish the configuration from APICs to switches, may be missing on the APICs after an upgrade. This is due to [CSCwp64296][64], introduced with the enhancement in 6.0(3).
```
lovkeshsharma702: we do not talk about the regression " This is due to [CSCwp64296][64], introduced with the enhancement in 6.0(3)." remove this.

lovkeshsharma702 (approved):

takishida:
## Example result outputs:

Pre-APIC Upgrade
```
[Check 85/91] Rogue EP Exception List missing on switches...                                                       FAIL - OUTAGE WARNING!!
  Rogue Exception MACs Count  presListener Count
  --------------------------  ------------------
  5                           N/A

  Recommended Action: Change the target version to a fixed version of CSCwp64296.
  Reference Document: https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#rogue-ep-exception-list-missing-on-switches
```


Post-APIC Upgrade
```
[Check 85/91] Rogue EP Exception List missing on switches...                                                       FAIL - OUTAGE WARNING!!
  Rogue Exception MACs Count  presListener Count
  --------------------------  ------------------
  5                           only 27 found out of 32

  Recommended Action: Delete the exception lists and create again before upgrading switches. Or contact Cisco TAC to restore the missing presListener objects.
  Reference Document: https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#rogue-ep-exception-list-missing-on-switches
```

PR #298, opened 2025-12-01
New validation for CFD CSCwn81692

muthu-ku:
Added validation for CFD CSCwn81692

Log:

 python aci-preupgrade-validation-script.py --tversion "6.1(2f)" -d "active_node_presListener_mo_object_check"
    ==== 2026-01-07T04-25-55+0000, Script Version v4.0.1  ====

!!!! Check https://github.com/datacenter/ACI-Pre-Upgrade-Validation-Script for Latest Release !!!!

To use a non-default Login Domain, enter apic#DOMAIN\\USERNAME
Enter username for APIC login          : admin
Enter password for corresponding User  :

Gathering Node Information...

Current APIC Version...6.0(9e)
Lowest Switch Version...4.2(3j)

Target APIC version is overridden to 6.1(2f)

Collecting VPC Node IDs...101, 103

Progress: |----------------------------------------------------------------------------------------------------| 0/1 checkProgress: |████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 checks completed

=== Check Result (failed only) ===

[Check 1/1] active_node pres.Listener mo object check... PresListener Object Missing Nodes: {'102'}                FAIL - OUTAGE WARNING!!
  Missing Node ID  Node Status
  ---------------  -----------
  102              active

  Recommended Action: Contact Cisco TAC to investigate missing leaf nodes from class-4307 preslisteners objects
  Reference Document: https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#active_node_presListener_mo_object_check

=== Summary Result ===

PASS                        :  0
FAIL - OUTAGE WARNING!!     :  1
FAIL - UPGRADE FAILURE!!    :  0
MANUAL CHECK REQUIRED       :  0
POST UPGRADE CHECK REQUIRED :  0
N/A                         :  0
ERROR !!                    :  0
TOTAL                       :  1

    Pre-Upgrade Check Complete.
    Next Steps: Address all checks flagged as FAIL, ERROR or MANUAL CHECK REQUIRED

    Result output and debug info saved to below bundle for later reference.
    Attach this bundle to Cisco TAC SRs opened to address the flagged checks.

      Result Bundle: /data/techsupport/preupgrade_validator_2026-01-07T04-25-55+0000.tgz

==== Script Version v4.0.1 FIN ====

Pytest log:
-------------

================================================== test session starts ===================================================
platform linux -- Python 3.8.10, pytest-8.3.5, pluggy-1.5.0 -- /usr/bin/python3
[...]

PR #339, opened 2026-02-04
issue NewValidation: CSCwn81692

### LINKED ISSUE #297: NewValidation: CSCwn81692
Validation Type
[ ] - Fault
[ ] - Config
[X] - Bug
[ ] - Other

**What needs to be validated**

CSCwn81692 - the condition can be found from the bug.

**Why it needs to be validated**

This check will verify all active node object presence in presListener mo list with class 4307 and alert if targeting an affected version. If alerted, the recommended action is to target a version that has the fix for this bug.

jeestr4d:
Fixes #297

### DIFF
--- a/aci-preupgrade-validation-script.py
+++ b/aci-preupgrade-validation-script.py
@@ -6032,6 +6032,7 @@ def pres_listener_mo_check(fabric_nodes, tversion, **kwargs):
     headers = ['Pod-ID', 'Node-ID', 'State']
     data = []
     recommended_action = 'Contact TAC to apply the workaround for the listed Nodes BEFORE Upgrade.'
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations#missing-preslistener-mo'
     # fabricNode.fabricSt shows `disabled` for both Decommissioned and Maintenance (GIR).
     # fabricRsDecommissionNode.debug==yes is required to show `disabled (Maintenance)`.
     presListener = icurl('class','presListener.json')
@@ -6065,7 +6066,7 @@ def pres_listener_mo_check(fabric_nodes, tversion, **kwargs):
     if not fabric_nodes:
         result = MANUAL
         msg = 'Active Switch Leaf fabricNode not found!'
-    return Result(result=result, msg=msg, headers=headers, data=data, recommended_action=recommended_action)
+    return Result(result=result, msg=msg, headers=headers, data=data, recommended_action=recommended_action, doc_url=doc_url)
 
 # ---- Script Execution ----
 
--- a/aci-preupgrade-validation-script.py
+++ b/aci-preupgrade-validation-script.py
@@ -6025,6 +6025,47 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
 
     return Result(result=result, headers=headers, data=data, recommended_action=recommended_action, doc_url=doc_url)
 
+@check_wrapper(check_title="presListener MO Status")
+def pres_listener_mo_check(fabric_nodes, tversion, **kwargs):
+    result = PASS
+    msg = ''
+    headers = ['Pod-ID', 'Node-ID', 'State']
+    data = []
+    recommended_action = 'Contact TAC to apply the workaround for the listed Nodes BEFORE Upgrade.'
+    # fabricNode.fabricSt shows `disabled` for both Decommissioned and Maintenance (GIR).
+    # fabricRsDecommissionNode.debug==yes is required to show `disabled (Maintenance)`.
+    presListener = icurl('class','presListener.json')
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+       
+    if tversion.newer_than("6.1(3f)"):
+        result = NA
+        return Result(result=result, msg=msg, headers=headers, data=data, recommended_action=recommended_action)
+    
+    for fabric_node in fabric_nodes:
+        found = False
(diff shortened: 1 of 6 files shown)


aci-preupgrade-validation-script.py
```diff
@@ -6025,6 +6025,48 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
...
+            continue
+        state = fabric_node['fabricNode']['attributes']['fabricSt']
+        if state != 'active':
+            continue
+        dn = re.search(node_regex, fabric_node['fabricNode']['attributes']['dn'])
+        pod_id = dn.group("pod")
+        node_id = dn.group("node")
+        for mo in presListener:
+            if (
+                node_id in mo['presListener']['attributes']['lstDn'] and
```
monrog2, line 6058: **Blocking:** This substring check treats leaf `101` as present when the only MO belongs to `1101`, producing a false `PASS` and potentially missing the documented port outage. Parse the terminal node ID or compare `lstDn` with the exact expected leaf path. Please add a 101/1101 collision test that expects `FAIL_O` and verifies the reported node.

aci-preupgrade-validation-script.py
```diff
@@ -6025,6 +6025,48 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
...
+            if (
+                node_id in mo['presListener']['attributes']['lstDn'] and 
+                'class-4307' in mo['presListener']['attributes']['dn']
+                ):
+                found = True
+                continue
+        if not found:
+            result = FAIL_O
+            data.append([pod_id, node_id, state])       
+    if not fabric_nodes:
```
monrog2, line 6066: **Blocking:** `fabric_nodes` normally includes controllers and spines, so a non-empty inventory with no active leaves skips every entry but still returns `PASS`. Track how many active leaves were actually evaluated and return `MANUAL` when that count is zero. Please cover controller-only and disabled-leaf inventories.

aci-preupgrade-validation-script.py
```diff
@@ -6025,6 +6025,48 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
...
+        return Result(result=result, msg=msg, headers=headers, data=data, recommended_action=recommended_action)
+    
+    for fabric_node in fabric_nodes:
+        found = False
+        if fabric_node['fabricNode']['attributes']['role'] != "leaf":
+            continue
+        state = fabric_node['fabricNode']['attributes']['fabricSt']
+        if state != 'active':
+            continue
+        dn = re.search(node_regex, fabric_node['fabricNode']['attributes']['dn'])
```
monrog2, line 6053: **High:** The regex result is dereferenced without validation. One malformed active-leaf DN raises an exception, changes the whole check to `ERROR`, and discards any missing-MO evidence already collected. Preserve valid findings and report malformed objects through the existing unformatted-data mechanism; add a mixed valid/malformed regression test.

aci-preupgrade-validation-script.py
```diff
@@ -6025,6 +6025,48 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
...
+def pres_listener_mo_check(fabric_nodes, tversion, **kwargs):
+    result = PASS
+    msg = ''
+    headers = ['Pod-ID', 'Node-ID', 'State']
+    data = []
+    recommended_action = 'Contact TAC to apply the workaround for the listed Nodes BEFORE Upgrade.'
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations#missing-preslistener-mo'
+    # fabricNode.fabricSt shows `disabled` for both Decommissioned and Maintenance (GIR).
+    # fabricRsDecommissionNode.debug==yes is required to show `disabled (Maintenance)`.
+    presListener = icurl('class','presListener.json')
```
monrog2, line 6038: **High:** This full-class query runs before the target-version guards. Therefore a missing or unaffected target can become `ERROR` if collection fails even though the result should be `MANUAL` or `N/A`, and every run unnecessarily retrieves all `presListener` objects. Move the guards before collection and use a server-side filter for the relevant class-4307/node entries. Please test that `icurl` is not called for missing or unaffected targets.

aci-preupgrade-validation-script.py
```diff
@@ -6025,6 +6025,48 @@ def apic_downgrade_compat_warning_check(cversion, tversion, **kwargs):
...
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+       
+    if tversion.newer_than("6.1(3f)"):
```
monrog2, line 6042: **High:** The exact release boundary is not documented or tested: this condition keeps `6.1(3f)` in the affected range, while the tests jump from `6.1(3a)` to `6.1(4a)`. Reconcile the threshold with CSCwn81692, document the affected/fixed releases, and add tests immediately below, exactly at, and immediately above the fixed boundary.

monrog2 (changes requested):
The new validation has multiple false-PASS paths and can discard outage evidence. Please address the inline findings and add the requested regression coverage before merging.

monrog2:
Closing as stale. This is a separate reimplementation of the CSCwn81692 validation previously proposed in #298. That earlier PR was closed by its author with the note that the validation would not be added to the pre-upgrade script based on the prior product discussion. This PR has had no implementation updates since February, is currently merge-conflicted, and still has five unresolved correctness blockers from the July review, including false-PASS and lost-evidence paths. @lovkeshsharma702, please ping @monrog2 if the decision changes. If revived, issue #297 should be updated with the current justification and the implementation must resolve the outstanding review blockers.

PR #355, opened 2026-02-25
Add arg to define thread limit - to throttle concurrent API calls when required

monrog2:
```
    parser.add_argument("--max-threads", action="store", type=int, default=None, help="Maximum number of check threads to run concurrently. Defaults to unlimited.")

### DIFF
--- a/tests/test_parse_args.py
+++ b/tests/test_parse_args.py
@@ -22,6 +22,7 @@ def test_no_args():
     assert args.no_cleanup is False
     assert args.version is False
     assert args.total_checks is False
+    assert args.max_threads is None
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: aci-preupgrade-validation-script.py [hunk 1]
```diff
@@ -1493,11 +1514,14 @@ def get_row(widths, values, spad="  ", lpad=""):
 
 def prints(objects, sep=' ', end='\n'):
     with open(RESULT_FILE, 'a') as f:
-        print(objects, sep=sep, end=end, file=sys.stdout)
+        try:
+            print(objects, sep=sep, end=end, file=sys.stdout)
+            sys.stdout.flush()
+        except OSError:
+            pass
```
monrog2: pytests were intermittently failing with `Bad File Descriptor`

tests/test_ThreadManager.py
```diff
@@ -9,41 +9,67 @@
...
+    # capsys.readouterr() is called.  Without this there is a race where a
+    # thread passes the `if not global_timeout` check and then tries to print
+    # after pytest has already torn down the captured stdout fd, causing
+    # OSError: [Errno 9] Bad file descriptor.
+    for thread in tm.threads:
+        try:
+            thread.join(timeout=1.5)
+        except RuntimeError:
+            pass  # thread was never started
+
```
monrog2: Handling another scenario where `Bad File descriptor` would intermittently fail this specific pytest

PR #393, opened 2026-05-31
CSCwu34390 CX Script: Added new rule to check if Fabric BFD is enable…

weinjason:
…d on ISIS

Fabric BFD (BFD on ISIS) is not recommended,. The operational benefit of enabling Fabric BFD is minimal to none, while the potential adverse impact is significantly higher. The key reasons are outlined below:

ISIS runs over directly connected sub-interfaces
In a leaf-spine fabric, ISIS peers are directly connected sub-interfaces. If a peer device (leaf or spine) goes down, the physical link also goes down, which triggers Layer-1 convergence. In such scenarios—representing the vast majority of failure cases—BFD does not provide any additional convergence benefit.

Risk of false BFD flaps:
BFD is susceptible to false flaps even when there is no actual network path issue. This can occur when BFD packets do not receive sufficient CPU cycles, especially during periods of high CPU utilization by other processes (for example, during tech-support collections). Such false BFD flaps directly impact ISIS adjacencies. Since ISIS is a critical control-plane protocol for the entire fabric, instability at this level can have a widespread and severe impact across the data-center network

Solution:
In pre-upgrade script, added new rule to check if Fabric BFD is enabled on ISIS. If it is enabled, fail the validation and recommend the user to disable it before upgrade or downgrade.


aci-preupgrade-validation-script.py
```diff
@@ -3756,6 +3756,25 @@ def isis_redis_metric_mpod_msite_check(**kwargs):
...
+def fabric_bfd_isis_check(**kwargs):
+    result = PASS
+    headers = ["L3 Interface Policy DN"]
+    data = []
+    recommended_action = 'Prior to upgrade or downgrade, disable Fabric BFD on ISIS.'
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#fabric-bfd-on-isis'
+
+    l3IfPols = icurl('class', 'l3IfPol.json')
+    for pol in l3IfPols:
+        if pol['l3IfPol']['attributes'].get('bfdIsis') == 'enabled':
```
kahande, line 3769: For no data found (l3IfPol completely missing), we will need to add manual validation. In case the l3IfPol is deleted with BFD set to enabled, the L3 uplink interfaces will continue to have BFD enabled on them. Test the behavior with custom L3 interface policy mapped to uplinks and deleting them to see what happens.

weinjason: I inspected the code. It shows that this behavior is expected by design.

weinjason: After the upgrade, the node performs a clean reload anyway. Because the policy is missing, BFD will be disabled on isisIf. Hence no extra handling needed in this case.

monrog2:
Closing as superseded by #401. Both PRs implement the same CSCwu34390 Fabric BFD validation from the same author and modify the same file set. The implementation and tests are identical; the only content difference is the documentation reference number required by the newer base. #401 targets the active `v4.2.0-dev` branch, while this PR targets `master`. Continue implementation and review in #401. A dedicated tracking issue still needs to be created and linked before #401 merges.

PR #403, opened 2026-06-30
Added validation check for CSCwt14573

### LINKED ISSUE #404: NewValidation: CSCwt14573
### Validation Type
[ ] - Fault

[ ] - Config

[X ] - Bug

[ ] - Other

1. VRF is stretched across multiple sites (has 
fvSiteAssociated
 with 
fvRemoteId
 children)
2. vzAny is used as either consumer OR provider on the stretched VRF
3. Contract has a service graph attached (any type - PBR is NOT required)
4. No 
vnsEpgDefXlate
 MOs exist for the service graph's first node consumer leg

Traffic impact if the service graph is detached and re-attached to the contract while the fault condition is present.

### Additional context
RC: CSCwn95571 added vzAny multisite validation that was previously missing

dhaselva:
[vzany_stretched_vrf_aci-run-logs_Jun30.txt](https://github.com/user-attachments/files/29502788/vzany_stretched_vrf_aci-run-logs_Jun30.txt)
[vzany_stretched_vrf_pytest-logs_Jun30.txt](https://github.com/user-attachments/files/29502790/vzany_stretched_vrf_pytest-logs_Jun30.txt)

Fixes #404


monrog2 (changes requested):
## Changes requested

### 1. Restrict detection to PBR service graphs

**File:** `aci-preupgrade-validation-script.py`, around line 7340

CSCwt14573 and CSCwn95571 require locally configured **PBR**. The current implementation processes every applied service graph, so a non-PBR graph without translation entries is incorrectly reported as `FAIL_O`.

Please filter graph instances using `vnsNodeInst.routingMode == "Redirect"`, consistent with the existing PBR detection in `consumer_vzany_shared_services_check`.

```python
graph_query = (
    'vnsGraphInst.json?'
    'query-target-filter=eq(vnsGraphInst.configSt,"applied")'
    '&rsp-subtree=full'
    '&rsp-subtree-class='
    'vnsNodeInst,vnsTermNodeInst,vnsConnectionInst,'
    'vnsRsConnectionInstConns'
)
graph_insts = icurl('class', graph_query)

for graph_inst_mo in graph_insts:
    graph_inst = graph_inst_mo.get('vnsGraphInst', {})
    children = graph_inst.get('children', [])

    is_pbr = any(
        child.get('vnsNodeInst', {})
        .get('attributes', {})
        .get('routingMode') == 'Redirect'
        for child in children
    )
    if not is_pbr:
        continue
```

Also update `docs/docs/validations.md` to remove the statements that PBR is not required.

Please add tests proving:

- `routingMode="unspecified"` returns `PASS`.
- `routingMode="Redirect"` with a missing translation returns `FAIL_O`.

### 2. Preserve every graph instance and stretched VRF

**File:** `aci-preupgrade-validation-script.py`, around lines 7383 and 7433

`sg_by_contract` and `vzany_on_stretched` are keyed only by contract DN. A contract can be used by multiple VRFs, and each scoped graph instance requires its own translation check. Later entries currently overwrite earlier ones.

I reproduced two stretched VRFs sharing one contract where the first translation was missing and the second existed. The check queried only the second instance and returned `PASS`.

Please retain all graph instances and identify relationships by both contract and VRF scope.

```python
sg_by_contract = defaultdict(list)

# While processing graph instances:
sg_by_contract[contract_dn].append({
    'gi_dn': gi_dn,
    'graph_name': graph_name,
    'scope_dn': scope_dn,
    'first_node': first_node_name,
})

vzany_on_stretched = set()

# While processing vzAny relationships:
vzany_on_stretched.add((contract_dn, vrf_dn))

# Evaluate every graph instance matching the relationship's VRF:
for contract_dn, vrf_dn in vzany_on_stretched:
    for sg_info in sg_by_contract.get(contract_dn, []):
        if sg_info['scope_dn'] != vrf_dn:
            continue

        # Check the translation for this specific graph instance.
```

Please add a regression test with two stretched VRFs using the same contract where only one graph instance has a translation. The result must identify the missing instance rather than return `PASS`.

### 3. Do not silently skip unparseable graph instances

**File:** `aci-preupgrade-validation-script.py`, around line 7440

When the first consumer node cannot be parsed, the check executes `continue`. If no other rows are generated, this produces a false `PASS`. A malformed contract DN is handled similarly.

Required parsing failures should produce `ERROR` and preserve the available graph evidence.

```python
unformatted_headers = ['Graph Instance DN', 'Issue']
unformatted_data = []

# ...

if not first_node_name:
    has_error = True
    unformatted_data.append([
        sg_info.get('gi_dn') or contract_dn,
        'Unable to determine the first consumer node',
    ])
    continue

contract_match = re.match(r'uni/tn-([^/]+)/brc-([^/]+)', contract_dn)
if not contract_match:
    has_error = True
    unformatted_data.append([
        sg_info.get('gi_dn') or contract_dn,
        'Unable to parse the contract DN',
    ])
    continue
```

Return the accumulated evidence:

```python
return Result(
    result=result,
    headers=headers,
    data=data,
    unformatted_headers=unformatted_headers,
    unformatted_data=unformatted_data,
    recommended_action=recommended_action,
    doc_url=doc_url,
)
```

Please add malformed/missing-node and malformed-contract tests asserting `ERROR` and the retained unformatted evidence.

### 4. Avoid retrieving every full service-graph subtree

**File:** `aci-preupgrade-validation-script.py`, around line 7340

The current request retrieves the complete subtree of every applied `vnsGraphInst` before establishing whether any relevant stretched VRF or vzAny relationship exists. This can create substantial APIC load on large fabrics.

Please collect stretched VRFs and relevant vzAny relationships first, then retrieve only the graph classes required for PBR and node discovery.

```python
graph_query = (
    'vnsGraphInst.json?'
    'query-target-filter=eq(vnsGraphInst.configSt,"applied")'
    '&rsp-subtree=full'
    '&rsp-subtree-class='
    'vnsNodeInst,vnsTermNodeInst,vnsConnectionInst,'
    'vnsRsConnectionInstConns'
)
```

The test query key must exactly match the optimized production query.

@lovkeshsharma702 Please review these requested changes, particularly the PBR trigger and multi-VRF handling, before the next revision.

aci-preupgrade-validation-script.py
```diff
@@ -7320,6 +7320,184 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+
+    if not tversion.newer_than("6.1(3g)"):
+        return Result(result=PASS, msg="Target version does not trigger CSCwn95571 validation", doc_url=doc_url)
+
+    has_error = False
+
+    # Pre-6.1(4): the impacted graph is still 'applied', so scope to that state to limit load.
+    # 6.1(4)+: a later re-render can leave it failed-to-apply, so poll all states.
+    graph_subtree = '&rsp-subtree=full&rsp-subtree-class=vnsTermNodeInst,vnsConnectionInst,vnsRsConnectionInstConns'
+    if cversion and not cversion.newer_than("6.1(3g)"):
```
monrog2: **[P1] Restrict this to PBR graphs.** This subtree does not request `vnsNodeInst`, and the implementation never checks `routingMode == "Redirect"`; therefore it evaluates every graph type. The existing changes request identifies CSCwt14573/CSCwn95571 as PBR-specific, so non-PBR graphs without translation entries can be reported incorrectly. Include `vnsNodeInst`, skip graphs without a Redirect node, and add Redirect/unspecified regression fixtures. Please also reconcile the docs, which currently say PBR is not required.

aci-preupgrade-validation-script.py
```diff
@@ -7320,6 +7320,184 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+        return Result(result=PASS, msg="No service graphs found", doc_url=doc_url)
+
+    # Key by contract with a list of instances so multiple VRFs sharing one contract are all retained.
+    sg_by_contract = defaultdict(list)
+    for graph_inst_mo in graph_insts:
+        gi_attrs = graph_inst_mo.get('vnsGraphInst', {}).get('attributes', {})
+        gi_dn = gi_attrs.get('dn', '')
+        contract_dn = gi_attrs.get('ctrctDn', '')
+        if not contract_dn:
+            continue
```
monrog2: **[P1] Do not silently discard a graph with no contract DN.** A required identifier missing from an APIC response means this graph cannot be evaluated; `continue` can leave the overall result as PASS and hide an affected configuration. Set `has_error`, retain `gi_dn` plus a descriptive issue in `unformatted_data`, and return ERROR. Add a regression fixture.

aci-preupgrade-validation-script.py
```diff
@@ -7320,6 +7320,184 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+
+        # Skip NDO/MSC-managed graphs; their translation entries are created by NDO.
+        annotation = gi_attrs.get('absGraphAnnotation', '') or gi_attrs.get('annotation', '')
+        if 'orchestrator:msc' in annotation:
+            continue
+
+        graph_name_match = re.search(r'-G-\[uni/tn-[^/]+/AbsGraph-([^\]]+)\]', gi_dn)
+        graph_name = graph_name_match.group(1) if graph_name_match else ''
+        scope_match = re.search(r'-S-\[(.*?)\]', gi_dn)
+        scope_dn = scope_match.group(1) if scope_match else ''
```
monrog2: **[P1] Treat an unparseable graph scope as an evaluation error.** When `-S-[...]` is absent or malformed, `scope_dn` becomes empty and the instance is silently skipped later because it cannot match the vzAny VRF. Record the graph DN and an explanatory unformatted error instead of allowing a false PASS; cover this with a test.

aci-preupgrade-validation-script.py
```diff
@@ -7320,6 +7320,184 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+            continue
+        for rel in rels:
+            rel_attrs = rel.get(rel_class, {}).get('attributes', {})
+            contract_dn = rel_attrs.get('tDn', '')
+            if contract_dn not in sg_by_contract:
+                continue
+            rel_dn = rel_attrs.get('dn', '')
+            vrf_dn_match = re.match(r'(uni/tn-[^/]+/ctx-[^/]+)', rel_dn)
+            if not vrf_dn_match:
+                continue
```
monrog2: **[P1] Do not ignore an unparseable vzAny relationship DN.** If a relevant relationship has a malformed DN, this path quietly drops it and may return PASS despite being unable to decide its VRF scope. Preserve the relationship DN and issue in `unformatted_data`, set `has_error`, and add a malformed-relation regression test.

tests/checks/vzany_svcgraph_stretched_vrf_check/test_vzany_svcgraph_stretched_vrf_check.py
```diff
@@ -0,0 +1,204 @@
...
+        (
+            {
+                vnsGraphInst_applied_query: read_data(dir, "vnsGraphInst_no_firstnode.json"),
+                fvCtx_query: read_data(dir, "fvCtx_stretched_vrf.json"),
+                vzRsAnyToCons_query: read_data(dir, "vzRsAnyToCons_consumer.json"),
+                vzRsAnyToProv_query: [],
+            },
+            "6.0(1a)",
+            "6.1(4a)",
+            script.ERROR,
```
monrog2: **[P2] Assert the diagnostic evidence, not only ERROR.** This case sets both expected data fields to `None`, so it verifies only the result code. Assert `unformatted_headers == ['Graph Instance DN', 'Issue']` and the retained DN/reason to prevent a regression that leaves operators with an unexplained ERROR. Add equivalent assertions for the malformed-DN cases.

monrog2 (changes requested):
Thanks for addressing the earlier multi-VRF retention and missing-consumer-node handling. I found the remaining issues below that should be resolved before merge.

Issue #386, opened 2026-05-19
NewValidation: check if Infra VLAN is present in some Domains that will fail the deploy on EPG from 6.1 release


palmierimarco:
(use upvote :thumbsup: for attentions)
### Validation Type
[ ] - Fault

[X] - Config

[ ] - Bug

[ ] - Other

### What needs to be validated
Starting from release 6.1 is not possible to bind a Physical Domain to a new EPG if that Domain has the Infra VLAN ID in the Vlan Pool range.
This script should check for this and notify the administrator to fix the vlan pools before the upgrade.

### Why it needs to be validated
The Infra VLAN ID must not be present inside any vlan pools, otherwise must be fixed before the upgrade.

monrog2:
Duplicate of #367. The active implementation is #405, which is currently under review. Consolidating tracking and discussion in #367.

monrog2:
Correction to the earlier duplicate assessment: authoritative defect metadata confirms that CSCwt58626 and CSCwn64461 are distinct defects. This issue tracks the InfraVLAN-in-user-VLAN-pool condition associated with CSCwt58626, which was implemented by #405 and released in v4.2.0. Leaving this issue closed as implemented; #367 remains the separate tracker for CSCwn64461.

PR #405, opened 2026-07-02
Added validation check for CSCwt58626

muthu-ku:
Added new validation for the defect CSCwt58626.

This check will check if the InfraVLAN overlaps with any user-configured VLAN pool range in Access Policies. 

Apic failure logs:

```
 python aci-preupgrade-validation-script.py --tversion "6.1(3g)" -d "infravlan_overlap_access_policy_check"
    ==== 2026-07-23T04-58-29+0000, Script Version v4.2.0-dev  ====

!!!! Check https://github.com/datacenter/ACI-Pre-Upgrade-Validation-Script for Latest Release !!!!

To use a non-default Login Domain, enter apic#DOMAIN\\USERNAME
Enter username for APIC login          : admin
Enter password for corresponding User  :

Gathering Node Information...

Current APIC Version...6.0(9e)
Lowest Switch Version...6.0(9e)

Target APIC version is overridden to 6.1(3g)

Collecting VPC Node IDs...101, 102, 103, 104

Progress: |----------------------------------------------------------------------------------------------------| 0/1 checks cProgress: |████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 checks completed


=== Check Result (failed only) ===

[Check 1/1] InfraVLAN Overlap in Access Policy VLAN Pools...                                                      FAIL - UPGRADE FAILURE!!
  InfraVLAN  Encap Block             VLAN Pool DN
  ---------  -----------             ------------
  3301       vlan-10 to vlan-4093    uni/infra/vlanns-[infra-vlan-pool]-static/from-[vlan-10]-to-[vlan-4093]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-akhila1/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-ashok/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-ashok2/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-burh414/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-dci_cluster/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-deep421/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-depro1/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-dualstack/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-experimentk8slab/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-fwardzic1/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-k8s_scale53/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-k8sdragan/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-k8sfwardzic/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-k8srkesetup1/from-[vlan-3301]-to-[vlan-3301]
  3301       vlan-3301 to vlan-3301  uni/vmmp-VMware/dom-k8s-scale-vmm/usrcustomaggr-k8srkesetup2/from-[vlan-3301]-to-[vlan-3301]

  Recommended Action: Remove InfraVLAN from VLAN pool block highligted or upgrade to fix version
  Reference Document: https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#infravlan-overlap-access-policy-check



=== Summary Result ===

PASS                        :  0
FAIL - OUTAGE WARNING!!     :  0
FAIL - UPGRADE FAILURE!!    :  1
MANUAL CHECK REQUIRED       :  0
POST UPGRADE CHECK REQUIRED :  0
N/A                         :  0
ERROR !!                    :  0
TOTAL                       :  1

    Pre-Upgrade Check Complete.
    Next Steps: Address all checks flagged as FAIL, ERROR or MANUAL CHECK REQUIRED

    Result output and debug info saved to below bundle for later reference.
    Attach this bundle to Cisco TAC SRs opened to address the flagged checks.

      Result Bundle: /data/techsupport/preupgrade_validator_2026-07-23T04-58-29+0000.tgz

==== Script Version v4.2.0-dev FIN ====


```
[apic_failure_log.txt](https://github.com/user-attachments/files/30293105/apic_failure_log.txt)
[apic_full_run_log.txt](https://github.com/user-attachments/files/30293108/apic_full_run_log.txt)
[pytest_full_run_log.txt](https://github.com/user-attachments/files/30293109/pytest_full_run_log.txt)

Implements #386


docs/docs/validations.md
```diff
@@ -2828,22 +2821,11 @@ To avoid this issue, disable WRED on the affected nodes or upgrade to a release
...
 
-
-### Stale dbgacEpgSummaryTask Objects
-
-Due to [CSCwt69100][75], a stale `dbgacEpgSummaryTask` object stuck in `processing` state with empty content can cause the policymgr process to crash on all APICs during an upgrade or process restart.
-
-Affected versions: 6.1(5e) and below, or 6.2(1g).
-
-Contact Cisco TAC for next steps. For more details, refer to the workaround in [CSCwt69100][75].
+To avoid this issue, modify user VLAN pool ranges so that the InfraVLAN does not overlap with any configured block or select the fixed version as a target release.
```
lovkeshsharma702: To avoid this issue, modify user VLAN pool ranges so that the InfraVLAN does not overlap with any configured block or select target version with fix.

tests/checks/infravlan_overlap_access_policy_check/lldpInst_infra_vlan.json
```diff
@@ -0,0 +1,10 @@
+[
```
lovkeshsharma702: add more info in lldpInst ouput like 6 entries.  there will be more for CU.

add failure check when lldp mo do not return anything. how do you handle this.?

aci-preupgrade-validation-script.py
```diff
@@ -6614,91 +6532,61 @@ def wred_affected_model_check(tversion, fabric_nodes, **kwargs):
...
-    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#n9k-c93180yc-fx3-switch-memory-less-than-32gb'
-    min_memory_kb = 32 * 1000 * 1000
-    msg = ''
-
-    affected_nodes = [
-        node for node in fabric_nodes
-        if node['fabricNode']['attributes']['model'] == 'N9K-C93180YC-FX3'
-    ]
+    recommended_action = (
+        "Ensure InfraVLAN is not part of any access VLAN pool range. "
```
lovkeshsharma702: "Remove InfraVLAN from VLAN pool block highligted or upgrade to fix version"

aci-preupgrade-validation-script.py
```diff
@@ -6614,91 +6532,61 @@ def wred_affected_model_check(tversion, fabric_nodes, **kwargs):
...
+        "Remove the overlapping InfraVLAN or adjust VLAN pool blocks."
```
lovkeshsharma702: remove this .

aci-preupgrade-validation-script.py
```diff
@@ -6614,91 +6532,61 @@ def wred_affected_model_check(tversion, fabric_nodes, **kwargs):
...
-        proc_mem_mos = icurl('class', query)
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
 
-        node_id_to_attrs = {
-            node['fabricNode']['attributes']['id']: node['fabricNode']['attributes']
-            for node in affected_nodes
-        }
+    if (tversion.same_as("6.2(1g)") or tversion.same_as("6.1(3f)") or
+        (tversion.newer_than("6.1(3f)") and tversion.older_than("6.1(5e)")) or tversion.same_as("6.1(5e)")):
```
lovkeshsharma702: better suited: 

"if (tversion.same_as("6.2(1g)") or
        (not tversion.older_than("6.1(3f)") and not tversion.newer_than("6.1(5e)"))):"

tests/checks/infravlan_overlap_access_policy_check/lldpInst_infra_vlan.json
lovkeshsharma702: add a case where you do not get any value in return for fvnsEncapBlk

aci-preupgrade-validation-script.py
```diff
@@ -6614,91 +6532,61 @@ def wred_affected_model_check(tversion, fabric_nodes, **kwargs):
...
-        dn = attr.get("dn", "")
-        start_ts = attr.get("startTs", "")
-        try:
-            task_dt = datetime.strptime(start_ts[:19], "%Y-%m-%dT%H:%M:%S")
-        except ValueError:
-            continue
-        if task_dt < threshold:
-            data.append([dn, start_ts])
+    else:
+        return Result(result=NA, msg=VER_NOT_AFFECTED)
```
lovkeshsharma702: Instead of early-returning NA for non-affected versions and keeping logic flat, it wraps all logic inside a positive block with [else: return NA] at the bottom. This adds unnecessary nesting and buries the [NA]

aci-preupgrade-validation-script.py
```diff
@@ -6614,91 +6532,61 @@ def wred_affected_model_check(tversion, fabric_nodes, **kwargs):
...
-            attrs = memory_mo['procMemUsage']['attributes']
-            dn_match = re.search(node_regex, attrs['dn'])
-            if not dn_match:
-                continue
-            node_id = dn_match.group('node')
-            if node_id not in node_id_to_attrs:
+        infra_vlan = None
+        lldpInsts = icurl('class', 'lldpInst.json')
+        for lldpInst in lldpInsts:
+            infra_vlan_id = lldpInst.get('lldpInst', {}).get('attributes', {}).get('infraVlan')
```
lovkeshsharma702: infra_vlan can be None — potential TypeError
If lldpInst returns no objects or none have infraVlan, infra_vlan stays None. Then min(from_vlan, to_vlan) <= infra_vlan raises a TypeError in Python 3. There's no guard before the encap_blocks loop:

aci-preupgrade-validation-script.py
```diff
@@ -6614,91 +6532,61 @@ def wred_affected_model_check(tversion, fabric_nodes, **kwargs):
...
+            try:
+                from_vlan = int(str(from_encap).split('-')[-1])
+                to_vlan = int(str(to_encap).split('-')[-1])
+            except (ValueError, TypeError):
+                continue
 
-    if tversion and ((tversion.major1 == "6" and tversion.major2 == "1" and tversion.newer_than("6.1(5e)")) or tversion.newer_than("6.2(1g)")):
-        return Result(result=NA, msg=VER_NOT_AFFECTED, doc_url=doc_url)
+            if min(from_vlan, to_vlan) <= infra_vlan <= max(from_vlan, to_vlan):
+                result = FAIL_O
```
lovkeshsharma702: # Current (inline, non-standard)
if min(from_vlan, to_vlan) <= infra_vlan <= max(from_vlan, to_vlan):
    result = FAIL_O     # ← set inside loop
    data.append([...])

# Standard
if data:
    result = FAIL_O     # ← set once after loop

lovkeshsharma702 (changes requested):

lovkeshsharma702 (approved):
all intigrated test passed.

docs/docs/validations.md
```diff
@@ -2846,6 +2848,13 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
 Contact Cisco TAC for next steps. For more details, refer to the workaround in [CSCwt69100][75].
 
 
+### Infravlan Overlap Access Policy Check
+
+Due to [CSCwt58626][77], when targeting APIC version 6.1(3f) to 6.1(5e) or 6.2(1g) , if the InfraVLAN overlaps with any user-configured VLAN pool range in Access Policies, fault `F4701` is raised for VLAN pools that include the InfraVLAN. After the upgrade, domains linked to those VLAN pools cannot be associated with new EPGs, although existing EPGs continue to function.
+
+To avoid this issue, modify user VLAN pool ranges so that the InfraVLAN does not overlap with any configured block or select target version with fix.
```
monrog2: @lovkeshsharma702 include some detail about the planned fix, otherwise this checks details will have to be updated again after release.

aci-preupgrade-validation-script.py
```diff
@@ -22,7 +22,7 @@
 from textwrap import TextWrapper
 from getpass import getpass
 from collections import defaultdict, OrderedDict
-from datetime import datetime
+from datetime import datetime, timedelta
```
monrog2, line 25: is timedelta used in this check?

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,65 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+
+    if not (tversion.same_as("6.2(1g)") or (
+        not tversion.older_than("6.1(3f)") and not tversion.newer_than("6.1(5e)")
+    )):
+        return Result(result=NA, msg=VER_NOT_AFFECTED)
+
+    infra_vlan = None
+    lldpInsts = icurl('class', 'lldpInst.json')
```
monrog2: don't need to query the full class if we only need 1/2 instances of hte object to pull infra vlan. risking scale issues.

Is there any concern with always getting node-1?

```shell
apic1:~> icurl 'http://127.0.0.1:7777/api//class/lldpInst.json?query-target-filter=wcard(lldpInst.dn,"node-1")'

{"totalCount":"1","imdata":[{"lldpInst":{"attributes":{...,"dn":"topology/pod-1/node-1/sys/lldp/inst","holdTime":"120","infraVlan":"3967"...}}]}

File: aci-preupgrade-validation-script.py [hunk 12]
```diff
@@ -6702,6 +6702,65 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+        infra_vlan_id = lldpInst.get('lldpInst', {}).get('attributes', {}).get('infraVlan')
+        if not infra_vlan_id:
+            continue
+        match = re.search(r'\d+', str(infra_vlan_id))
+        if match:
+            infra_vlan = int(match.group(0))
+            break
+
+    if infra_vlan is None:
+        return Result(result=NA, msg="Unable to determine InfraVLAN from lldpInst.")
```
monrog2: is NA the right response result if we fail to pull an object? maybe MANUAL makes more sense, or ERR if inability to pull this object is never expected.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,65 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+            from_vlan = int(str(from_encap).split('-')[-1])
+            to_vlan = int(str(to_encap).split('-')[-1])
+        except (ValueError, TypeError):
+            continue
+
+        if min(from_vlan, to_vlan) <= infra_vlan <= max(from_vlan, to_vlan):
+            data.append([str(infra_vlan), pool_name, "{} to {}".format(from_encap, to_encap)])
+
+    if data:
+        result = FAIL_O
```
monrog2: Does this cause an outage? it sounds like changes are blocked, but does it prevent programming post-upgrade to an affected version?

monrog2 (changes requested):

aci-preupgrade-validation-script.py
lovkeshsharma702: > is timedelta used in this check?
@monrog2 
dbgacEpgSummaryTask using this timedelta, it was causing issue on 5.2 code due to old python version, needed timedelta specification to handle this in old python code running on 5.2.  earlier intigrated test passsed i guess that time all lab were on 6.x.  now some of these downgraded and failed the test.

docs/docs/validations.md
lovkeshsharma702: > @lovkeshsharma702 include some detail about the planned fix, otherwise this checks details will have to be updated again after release.

@monrog2  this was a regression in 6.1(3f) to 6.1(5e) or 6.2(1g). fixed above 615 and 6.2.x. only 6.2.2eis release so far, hence fix version for 615 not mentioned in the bug. @muthu-ku please add 6.2.2e as fixed version in bug.

aci-preupgrade-validation-script.py
lovkeshsharma702: > don't need to query the full class if we only need 1/2 instances of hte object to pull infra vlan. risking scale issues.
> 
> Is there any concern with always getting node-1?
> 
> ```shell
> apic1:~> icurl 'http://127.0.0.1:7777/api//class/lldpInst.json?query-target-filter=wcard(lldpInst.dn,"node-1")'
> 
> {"totalCount":"1","imdata":[{"lldpInst":{"attributes":{...,"dn":"topology/pod-1/node-1/sys/lldp/inst","holdTime":"120","infraVlan":"3967"...}}]}
> ```

@muthu-ku please change this as discussed for Node1 as this is surely present.

aci-preupgrade-validation-script.py
lovkeshsharma702: > is NA the right response result if we fail to pull an object? maybe MANUAL makes more sense, or ERR if inability to pull this object is never expected.

@monrog2  manual used mainly once we conclude something and want cu to doublecheck, i guess in this case, it's more of fault handling otherwise infra_vlan has to be there..

aci-preupgrade-validation-script.py
lovkeshsharma702: > Does this cause an outage? it sounds like changes are blocked, but does it prevent programming post-upgrade to an affected version?

@monrog2  i was little doubtfull about this. but this is serious thing and cu once hit looking for a SMU as solution . already hit by 4 cu. we want this to be taken care by CU for sure hence highlighting it in FAIL_O.  however the issue can only block new EPG deployment if they used Vlanpool with infra vlan in block.

aci-preupgrade-validation-script.py
muthu-ku: Updated the required changes

aci-preupgrade-validation-script.py
monrog2: FAIL_UF is fine here

aci-preupgrade-validation-script.py
monrog2: use ERR result since we do not expect object queries to return empty

aci-preupgrade-validation-script.py
monrog2: ack, this is needed for another check

docs/docs/validations.md
monrog2: Main point of this one is to call out the behavior change implemented in the fixed versions. 

my understanding is:
PD will block path binding on user tenant if the binding it pushing the infra VLAN. 

HOWEVER, Infra tenant can do this without issue. 

affected version behavior relies solely on access policy lookup, which means whether infra or user tenant, any extension of infra VLAN causes the fault + blocking condition seen today.

point for clarification is how this would affected vmm bindings for infra vlan within user tenant, such as k8 integration

aci-preupgrade-validation-script.py
muthu-ku: upadated comment

docs/docs/validations.md
lovkeshsharma702: @monrog2  check with Dev,  they confirm fault raised if infra vlan found in any vlanpool in any tenant. we suggest cu not to use infra vlan in any of vlan pool, or move to fix version, where no fault is being raised.  
@muthu-ku please update the "Due to [CSCwt58626][77], when targeting APIC version 6.1(3f) to 6.1(5e) or 6.2(1g) , if the InfraVLAN overlaps with any VLAN pool range in Access Policies, fault `F4701` is raised for VLAN pools that include the InfraVLAN. After the upgrade, domains linked to those VLAN pools cannot be associated with new EPGs, although existing EPGs continue to function.

To avoid this issue, modify user VLAN pool ranges so that the InfraVLAN does not overlap with any configured vlan block or select target version with fix."

docs/docs/validations.md
```diff
@@ -2846,6 +2848,15 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
 Contact Cisco TAC for next steps. For more details, refer to the workaround in [CSCwt69100][75].
 
 
+### Infravlan Overlap Access Policy Check
+
+Due to [CSCwt58626][77], when targeting APIC version 6.1(3f) to 6.1(5e) or 6.2(1g) , if the InfraVLAN overlaps with any user-configured VLAN pool range in Access Policies, fault `F4701` is raised for VLAN pools that include the InfraVLAN. After the upgrade, domains linked to those VLAN pools cannot be associated with new EPGs, although existing EPGs continue to function.
+
+To avoid this issue, modify user VLAN pool ranges so that the InfraVLAN does not overlap with any configured block or select non-impacted version.
```
lovkeshsharma702: To avoid this issue, modify user VLAN pool ranges so that the InfraVLAN does not overlap with any configured block or select non-impacted/fixed version.

aci-preupgrade-validation-script.py
```diff
@@ -6685,7 +6685,12 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
     if tversion and ((tversion.major1 == "6" and tversion.major2 == "1" and tversion.newer_than("6.1(5e)")) or tversion.newer_than("6.2(1g)")):
         return Result(result=NA, msg=VER_NOT_AFFECTED, doc_url=doc_url)
 
-    threshold = datetime.utcnow() - timedelta(hours=24)
+    try:
+        from datetime import timezone 
+        threshold = datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(hours=24) 
+    except ImportError:
+        threshold = datetime.utcnow() - timedelta(hours=24)
```
lovkeshsharma702, line 6707: this potion is needed to cover 
~~~
 /data/techsupport/muthu/aci-preupgrade-validation-script.py:6688: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
threshold = datetime.utcnow() - timedelta(hours=24)

Python version	          Path taken	                                                                              Result
Python 2.x	          ImportError caught → datetime.utcnow()	                             ✅ Works
Python 3.0–3.11.        timezone import succeeds → datetime.now(timezone.utc)	✅ Works, no warning
Python 3.12+	         timezone import succeeds → datetime.now(timezone.utc)	✅ Works, no deprecation warning

lovkeshsharma702 (changes requested):

lovkeshsharma702 (approved):

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6707,65 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+    data = []
+    recommended_action = "Remove InfraVLAN from VLAN pool block highligted or upgrade to fix version"
+    
+    doc_url = "https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#infravlan-overlap-access-policy-check"
+
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+
+    if not (tversion.same_as("6.2(1g)") or (
+        not tversion.older_than("6.1(3f)") and not tversion.newer_than("6.1(5e)")
```
monrog2, line 6742: **[P1] Align the affected-version gate with authoritative defect data.** The current CSCwt58626 CDETS release note states the condition as `6.1.3f+ or 6.2.1g`, while this gate returns `N/A` for `6.1(6a)` and later. Please either extend the gate to match the current condition or cite and encode authoritative fixed-release metadata that establishes `6.1(5e)` as the upper bound. Otherwise, later affected 6.1 upgrades can be missed.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6707,65 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+            infra_vlan = int(match.group(0))
+            break
+
+    if infra_vlan is None:
+        return Result(result=ERROR, msg="Unable to determine InfraVLAN from lldpInst.")
+
+    encap_blocks = icurl('class', 'fvnsEncapBlk.json')
+    dn_pool_re = re.compile(r'vlanns-\[(?P<vlan_pool>[^\]]+)\]')
+    for obj in encap_blocks:
+        blk_attr = obj.get('fvnsEncapBlk', {}).get('attributes', {})
```
monrog2, line 6767: **[P1] Filter out non-affected internal blocks.** The unfiltered `fvnsEncapBlk` class response also contains blocks with `role: internal`, but this loop treats every block as a user VLAN-pool block. A valid internal block containing the InfraVLAN therefore produces `FAIL_UF`; I reproduced that behavior with an existing internal-role fixture shape. The CDETS example is `role: external`. Please restrict evaluation to the affected role/object types, and add an internal-role regression case.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6707,65 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+            continue
+
+        pool_match = dn_pool_re.search(dn)
+        pool_name = pool_match.group('vlan_pool') if pool_match else '-'
+
+        try:
+            from_vlan = int(str(from_encap).split('-')[-1])
+            to_vlan = int(str(to_encap).split('-')[-1])
+        except (ValueError, TypeError):
+            continue
```
monrog2: **[P1] Do not turn unparseable APIC data into a PASS.** Missing `dn`/`from`/`to` values and malformed VLAN values are silently skipped. If every returned block is malformed, `data` remains empty and the check reports `PASS` without establishing that no overlap exists; this is reproducible with a malformed-only response. Track these objects via `unformatted_data` and return `ERROR` (or the repository's equivalent incomplete-data result), with malformed-only and mixed valid/malformed tests.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6707,65 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+    for obj in encap_blocks:
+        blk_attr = obj.get('fvnsEncapBlk', {}).get('attributes', {})
+        dn = blk_attr.get('dn', '')
+        from_encap = blk_attr.get('from')
+        to_encap = blk_attr.get('to')
+        if not dn or not from_encap or not to_encap:
+            continue
+
+        pool_match = dn_pool_re.search(dn)
+        pool_name = pool_match.group('vlan_pool') if pool_match else '-'
```
monrog2: **[P2] Preserve an actionable identifier for VMM objects.** The canonical CDETS object DN is `uni/vmmp-VMware/dom-.../usrcustomaggr-.../from-[vlan-...]-to-[vlan-...]`, which has no `vlanns-[...]` component. This code detects the overlap but renders the pool as `-`, so the operator cannot locate the object named by the recommended action. Use the full DN (or a parsed VMM domain/custom aggregate identifier) as the fallback instead of `-`.

docs/docs/validations.md
```diff
@@ -2846,6 +2848,15 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
 Contact Cisco TAC for next steps. For more details, refer to the workaround in [CSCwt69100][75].
 
 
+### Infravlan Overlap Access Policy Check
+
+Due to [CSCwt58626][77], when targeting APIC version 6.1(3f) to 6.1(5e) or 6.2(1g) , if the InfraVLAN overlaps with any user-configured VLAN pool range in Access Policies, fault `F4701` is raised for VLAN pools that include the InfraVLAN. After the upgrade, domains linked to those VLAN pools cannot be associated with new EPGs, although existing EPGs continue to function.
```
monrog2: **[P2] Document the primary CDETS configuration, not only Access Policy pools.** CSCwt58626 specifically reports Kubernetes and VMware VMM custom port groups generated by `acc-provision`, including `usrcustomaggr` objects annotated `orchestrator:aci-containers-controller` with `containsInfraVlan: yes`. Describing the scope only as user-configured VLAN pool ranges in Access Policies can lead affected VMM/Kubernetes users to dismiss the warning. Please include these orchestrator-created objects and their post-upgrade impact.

tests/checks/infravlan_overlap_access_policy_check/test_infravlan_overlap_access_policy_check.py
```diff
@@ -0,0 +1,148 @@
...
+			},
+			"6.2(1g)",
+			script.ERROR,
+			[],
+		),
+		# Case 6: InfraVLAN overlaps on lower boundary version 6.1(3f). Expected: FAIL_UF.
+		(
+			{
+				lldpInst_api: read_data(dir, "lldpInst_infra_vlan_multiple_entry.json"),
+				fvnsEncapBlk_api: read_data(dir, "fvnsEncapBlk_overlap_single_vlan_pool.json"),
```
monrog2, line 69: **[P2] Add coverage using the actual CDETS object shape.** The overlap fixtures only use synthetic `vlanns-[pool]` DNs and omit `role`, `containsInfraVlan`, and `annotation`, so they cannot catch the internal-role false positive or the unusable `-` identifier for VMM `usrcustomaggr` objects. Add a fixture based on the CDETS example, plus internal-role and malformed-object cases, and assert the displayed identifier as well as the result.

monrog2 (changes requested):
Requesting changes for correctness against the current CSCwt58626 CDETS data and for fail-safe handling of APIC responses. Each concern is attached inline.

aci-preupgrade-validation-script.py
lovkeshsharma702: @monrog2  yes bug should be updated with 616x once available .bug is new and yet to be updated with 616 version, we can't add that id as it;s yet to be pasted in CCO. hence keeping tversion.newer_than("6.1(5e)") is the best way possible.

aci-preupgrade-validation-script.py
lovkeshsharma702: @monrog2  copy that.   @muthu-ku  please rework this to fetch the external block only.

aci-preupgrade-validation-script.py
lovkeshsharma702: @muthu-ku  please  work on the suggestions.

aci-preupgrade-validation-script.py
lovkeshsharma702: @monrog2  do you have a vmm setup where i can check.  i think when i checked in QA< i found a corresponding  vlan pool created with "vlanns-" in db, hence suggested to ignore.   but let me double check on this.

docs/docs/validations.md
lovkeshsharma702: copy that.  @muthu-ku  please update the bug and script document as well.

tests/checks/infravlan_overlap_access_policy_check/test_infravlan_overlap_access_policy_check.py
lovkeshsharma702: @muthu-ku  lets see if can get the rn and role from the moquery to compare and whereever you find the conflicing Mo, report that whole DN .

lovkeshsharma702 (approved):

aci-preupgrade-validation-script.py
lovkeshsharma702: when you find unformatted Dn, exit loop and print error.

lovkeshsharma702 (changes requested):

monrog2:
@lovkeshsharma702 try this way

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6713,81 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+    recommended_action = "Remove InfraVLAN from VLAN pool block highligted or upgrade to fix version"
+    
+    doc_url = "https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#infravlan-overlap-access-policy-check"
+
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+
+    if not (tversion.same_as("6.2(1g)") or (
+        not tversion.older_than("6.1(3f)") and not tversion.newer_than("6.1(5e)")
+    )):
```
monrog2, line 6743: **[P1] Do not stop checking 6.1 targets after 6.1(5e).** The current CSCwt58626 CDETS release note explicitly lists the condition as `Running ACI version 6.1.3f+ or 6.2.1g`. This upper bound therefore returns `N/A` for later affected 6.1 targets such as 6.1(6a). CDETS has internal integration builds (`6.1(5.10)`, `6.2(1.137)`, `6.2(2.139)`, and `6.3(0.90)`), but those do not establish a customer-facing fixed 6.1 release. Please gate all 6.1 versions from 6.1(3f) onward until an authoritative customer release is documented, and add a later-6.1 regression test.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6713,81 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+                data.append(row)
+            else:
+                unformatted_data.append(row)
+
+    if not data and not unformatted_data:
+        result = PASS
+
+    if has_error or unformatted_data:
+        result = ERROR
+        msg = "Overlap check for InfraVLAN {} could not be determined because one or more VLAN pool blocks contain invalid or improperly formatted data.".format(infra_vlan)
```
monrog2: **[P1] Preserve `FAIL_UF` whenever an overlap has already been confirmed.** This unconditional assignment downgrades a known upgrade-failure condition to `ERROR` if a later block is malformed, and it also converts overlaps stored in `unformatted_data` to `ERROR`. In the built-in ACI result mapping that changes severity from critical to major. Keep `FAIL_UF` when either `data` or `unformatted_data` proves an overlap; use `ERROR` only when malformed input prevents any conclusion. Please update the malformed-DN expectation and add a mixed confirmed-overlap/malformed-object test.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6713,81 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+        if not dn or not from_encap or not to_encap:
+            has_error = True
+            break
+
+        try:
+            from_vlan = int(str(from_encap).split('-')[-1])
+            to_vlan = int(str(to_encap).split('-')[-1])
+        except (ValueError, TypeError):
+            has_error = True
+            break
```
monrog2: **[P1] Continue scanning after malformed blocks.** Both error paths `break` out of the class response, so a malformed object near the beginning prevents later valid overlaps from being detected or shown. I reproduced this with a missing-DN object followed by a valid overlapping external block: the result contains no actionable overlap row. Track the response as incomplete, preserve whatever identifiers are available in `unformatted_data`, and continue evaluating the remaining blocks. Add a regression case where malformed data precedes a valid overlap.

docs/docs/validations.md
```diff
@@ -2846,6 +2848,12 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
 Contact Cisco TAC for next steps. For more details, refer to the workaround in [CSCwt69100][75].
 
 
+### Infravlan Overlap Access Policy Check
+
+Due to the bug [CSCwt58626][77] , If Apic upgrade planned for target versions 6.1(3f), 6.1(3g), 6.1(4h), 6.1(5e) and 6.2(1g), be aware of fault F4701 being raised if the InfraVLAN overlaps with any user-configured VLAN pool in Access Policies. This also affects vlan pool created by NDO/MSO, VMM, Kubernetes. After the upgrade, domains associated with those VLAN pools cannot be linked to new EPGs, although existing EPGs continue to function.
```
monrog2, line 2857: **[P2] Align the documented affected range with the current CDETS release note.** CDETS now states `6.1.3f+ or 6.2.1g`, while this sentence enumerates selected 6.1 releases and implies that unlisted versions such as 6.1(4a) or later 6.1 builds are not affected. Document the 6.1(3f)-and-later condition (and exact 6.2(1g) condition), and only name a fixed customer release once authoritative release metadata establishes it.

monrog2 (changes requested):
Requesting changes for the remaining CSCwt58626 correctness issues. The current CDETS release note now explicitly states the affected condition as `6.1.3f+ or 6.2.1g`; the inline comments also cover preserving confirmed upgrade-failure evidence and completing inventory evaluation.

lovkeshsharma702:
@monrog2  please review the PR again. commented changes were completed and tested with successful integration test.

aci-preupgrade-validation-script.py
monrog2: Withdrawing this requested change after checking CDETS integration metadata. CSCwt58626 was found in internal build 6.1(5.6) (the RNE customer case is 6.1(5e)) and is integrated in 6.1(5+). The expected assumption is that the integrated 6.1(5) fix carries into the future 6.1(6) baseline, so treating versions after 6.1(5e) as unaffected is reasonable. No change requested for this gate.

docs/docs/validations.md
monrog2: Withdrawing this documentation request as well. Given the CDETS integration in 6.1(5.10) and the assumption that future 6.1(6) releases inherit that fix, listing the currently published affected targets through 6.1(5e) is consistent with the intended scope.

aci-preupgrade-validation-script.py
lovkeshsharma702: @muthu-ku  can you revisit the change and correct is as per request.

aci-preupgrade-validation-script.py
lovkeshsharma702: @muthu-ku  i rememeber having this discussion with you and suggested the change, lets change it to better option as suggested.

docs/docs/validations.md
lovkeshsharma702: updated bug release notes.  @muthu-ku  mark it resolved.

lovkeshsharma702 (approved):

lovkeshsharma702:
@monrog2 please review the PR.  commented part is taken care of.

aci-preupgrade-validation-script.py
muthu-ku: Updated the script to continue evaluating incomplete response (unformatted_data) along with valid overalpped data. removed break statement.

aci-preupgrade-validation-script.py
muthu-ku: updated script to raise FAIL_UF when either data or unformatted_data proves an overlap. result will be error when there is no overlapped entry and found any missing parameterr in response.

docs/docs/validations.md
muthu-ku: Removed the check to capture only user configured vlan vlanns-[...] ,, now script wil capture all Dns with overalpped entry for  VMM/Kubernetes users  as well.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6713,81 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+        return Result(result=ERROR, msg="Unable to determine InfraVLAN from lldpInst.")
+
+    encap_blocks = icurl('class', 'fvnsEncapBlk.json?query-target-filter=eq(fvnsEncapBlk.role,"external")')
+    for obj in encap_blocks:
+        blk_attr = obj.get('fvnsEncapBlk', {}).get('attributes', {})
+        dn = blk_attr.get('dn', '')
+        from_encap = blk_attr.get('from')
+        to_encap = blk_attr.get('to')
+
+        if not dn or not from_encap or not to_encap:
```
monrog2, line 6773: **[P1] Preserve incomplete overlapping blocks in the failure evidence.** This path sets `has_error` and drops the object entirely. Case 16 already includes a missing-DN block whose `from`/`to` range contains the InfraVLAN, but the test expects no `unformatted_data`; because a later valid overlap keeps the overall result at `FAIL_UF`, the operator sees neither the omitted affected block nor the incomplete-inventory warning added by `AciResult`. That can lead them to remediate only the displayed pool and leave another upgrade-impacting overlap. When enough range information exists to establish overlap, preserve the available identifier (`dn` or `rn`, with a placeholder if necessary) in `unformatted_data`, continue scanning, and update Case 16 to assert that evidence and warning.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6713,81 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+@check_wrapper(check_title="InfraVLAN Overlap in Access Policy VLAN Pools")
+def infravlan_overlap_access_policy_check(tversion, **kwargs):
+    result = FAIL_UF
+    msg = ""
+    headers = ["InfraVLAN", "Encap Block", "VLAN Pool DN"]
+    unformatted_headers = ["InfraVLAN", "Encap Block", "VLAN Pool DN"]
+
+    data = []
+    unformatted_data = []
+    recommended_action = "Remove InfraVLAN from VLAN pool block highligted or upgrade to fix version"
```
monrog2: **[P2] Do not prescribe unsupported edits to orchestrator-managed blocks.** The current CSCwt58626 defect record lists `Workaround: NA`, and this check intentionally includes VMM/Kubernetes blocks generated by `acc-provision`. Telling users to remove the InfraVLAN from those blocks may direct them to make unsupported changes that are regenerated by the orchestrator. Please align this guidance with the authoritative record—for example, select a non-affected target or contact Cisco TAC for remediation guidance—and add coverage for the emitted `recommended_action` and `doc_url`.

monrog2 (changes requested):
Requesting changes for the two remaining CSCwt58626 issues: preserve incomplete overlapping inventory evidence, and align remediation guidance with the current defect record.

aci-preupgrade-validation-script.py
monrog2: RNE should be updated with proper workaround steps we are proposing here

aci-preupgrade-validation-script.py
muthu-ku: updated the script for missing  dn/from/to and marked as error.  And if these malformed data has overlapped entry , captured in unformatted data and returned with fail_uf even marked error due to some missing dn.

aci-preupgrade-validation-script.py
muthu-ku: Updated the script with a requested change. whenever we have data (from/to) to identify overlapped entry,  we are checking whether overlapped or not. 

in this case if overlapped entry and rn found, dn also matched found - adding this into data else, can be any missing/malformed dn so always added in unformatted data. 

when no from/to or malformed vlan found. marked as Error. But when found overlapped entry in further iteration then we are adding anyway data/unsupported data, result will be Fail_UF else result remains Error.

For missing dn overlapped entry, added additional available identifier in unformatted data. Other than rn.. no meaningful data present in the response.

lovkeshsharma702 (approved):

lovkeshsharma702 (approved):

aci-preupgrade-validation-script.py
lovkeshsharma702: @monrog2  followed the bug release note and removed WA from the script.

aci-preupgrade-validation-script.py
lovkeshsharma702: @monrog2  Rn we removed as in one of initigrated test we don't see rn attribute coming in the Icurl.  modified the logic accordingly.

lovkeshsharma702 (approved):

lovkeshsharma702 (approved):

lovkeshsharma702:
@monrog2  lets sync up today if anymore changes neede on this PR.

monrog2:
Integration test summary (identifying environment details omitted):

- Six version-pair integration jobs ran against 22 configured environments.
- All six jobs completed but failed overall because of existing validation findings, authentication/permission failures, or unrelated script errors.
- 32 runs completed without a script `ERROR`.
- The InfraVLAN overlap check produced no `ERROR` outcomes.
- Three InfraVLAN upgrade-blocking findings were detected across two environments, limited to the `6.1(4h)` target. Environment names, VLAN IDs, and policy paths are intentionally omitted.

| Version path | Clean completions | Timing range | Average |
|---|---:|---:|---:|
| `5.2(4d) -> 5.2(8f)` | 6 | 16-30s | 22s |
| `5.2(4d) -> 6.0(2h)` | 6 | 15-26s | 18s |
| `5.2(6e) -> 6.0(5a)` | 6 | 15-21s | 18s |
| `5.2(8d) -> 6.1(1f)` | 6 | 12-31s | 18s |
| `5.3(2d) -> 6.1(4h)` | 4 | 15-21s | 18s |
| `6.0(2a) -> 6.1(4h)` | 4 | 15-21s | 18s |

Overall clean-run timing was **12-31 seconds**, averaging approximately **19 seconds**.

Pytest validation also passed on both supported runtimes: **940 tests passed** on Python 2.7 and Python 3.8, including **18/18 parametrized InfraVLAN overlap cases** on each runtime.

PR #407, opened 2026-07-17
Added check for certificate expiration issue

### LINKED ISSUE #408: NewValidation: APIC SSL cert expire validation
### Validation Type
[x] - Fault

[ ] - Config

[ ] - Bug

[ ] - Other

**Expiring Certificates (Major Severity):**

* **F4501**: KeyRing X.509 Certificate expiring - This fault occurs when a custom KeyRing X.509 Certificate is going to expire in one month.

* **F3081**: SAML X.509 Certificate expiring - This fault occurs when the SAML X.509 Certificate is going to expire in one month.

* **F4617**: TP X.509 Certificate expiring - This fault occurs when a Trust Point X.509 Certificate is expiring.

* **F4752**: Factory X.509 Certificate expiring - This fault occurs when the factory Certificate is expiring.

**Expired Certificates (Critical Severity):**

* **F4502**: KeyRing X.509 Certificate expired - This fault occurs when a custom KeyRing X.509 Certificate has expired.

* **F4503**: TP X.509 Certificate expired - This fault occurs when a Trust Point X.509 Certificate has expired.

* **F3082**: SAML X.509 Certificate expired - This fault occurs when the SAML Encryption X.509 Certificate has expired.

* **F4753**: Factory X.509 Certificate expired - This fault occurs when the factory Certificate has expired.

asraf-khan:
Detection logic:

1. Find the fault using below query 
       'faultInst.json?query-target-filter=or(eq(faultInst.code,"F4501"),eq(faultInst.code,"F4502"),eq(faultInst.code,"F4503"),eq(faultInst.code,"F3081"),eq(faultInst.code,"F3082"),eq(faultInst.code,"F4617"),eq(faultInst.code,"F4752"),eq(faultInst.code,"F4753"))'
 
2. if one fault F4501, F3081, F4617, F4752 found and state is “raised” or “soaking”
         Result = Manual & recommended_action ="Renew the certificate(s) before it expires to avoid service disruption."
 
3. if one fault F4502, F4503, F3082, F4753 and state is “raised” or “soaking”
          Result = FAIL_O & recommended_action="Renew the certificate(s) immediately to restore functionality."
 
4. if faults found as combination (e.g. expiring F3081 & expired F4502 ) means 
          Result = FAIL_O & recommended_action="Renew expired certificate(s) immediately. For certificate(s) approaching expiry, renew before they expire to avoid service disruption."
          
 
 Faults            Short Description                             Introduced         First build
F3081, F3082      SAML encryption cert expiring/expired     2018-02-22        3.1(2f)
F4501, F4502      KeyRing cert expiring/expired             2023-09-13        6.0(4c)
F4503, F4617      TP cert expired/expiring                   2024-04-22        6.1(1e)
F4752,F4753              Factory cert expiring/expired                                                            6.1(5e)

For Factory certificate status if cversion < 6.1(5e), check via cli command "acidiag verifyapic"

Fixes #408

### DIFF
--- a/aci-preupgrade-validation-script.py
+++ b/aci-preupgrade-validation-script.py
@@ -6891,8 +6891,7 @@ def certificate_expiration_check(cversion, username, password, fabric_nodes, **k
             recommended_action = (
                     "Manually verify certificate expiry for all affected nodes. "
                     "\n\tFor APIC factory certificates, run `acidiag verifyapic` on each affected APIC. "
-                    "\n\tFor leaf/spine certificates, validate 'moquery -c pkiFabricNodeSSLCertificate' and check 'validityNotAfter' values via APIC API "
-                    "and resolve any connectivity or data parsing issues before upgrade."
+                    "\n\tFor leaf/spine certificates, validate 'moquery -c pkiFabricNodeSSLCertificate' and check 'validityNotAfter' values for affected nodes."
                 )
         elif has_critical and has_major:
             result = FAIL_O
--- a/aci-preupgrade-validation-script.py
+++ b/aci-preupgrade-validation-script.py
@@ -6702,6 +6702,116 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
     return Result(result=result, headers=headers, data=data, recommended_action=recommended_action, doc_url=doc_url)
 
 
+@check_wrapper(check_title='Certificate Expiration Check')
+def certificate_expiration_check(cversion, username, password, fabric_nodes, **kwargs):
+    result = PASS
+    headers = ["Fault Code", "Severity", "Description"]
+    data = []
+    recommended_action = ""
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#certificate-expiration-check'
+
+    fault_min_versions = {
+        "F4501": "6.0(4c)", "F4502": "6.0(4c)",  # KeyRing cert expiring/expired
+        "F4503": "6.1(1e)", "F4617": "6.1(1e)",  # TP cert expired/expiring
+        "F3081": "3.1(2f)", "F3082": "3.1(2f)",  # SAML encryption cert expiring/expired
+        "F4752": "6.1(5e)", "F4753": "6.1(5e)",  # Factory certificate expired/expiring
+    }
+    FACTORY_CERT_MIN_VERSION = "6.1(5e)"
+    FACTORY_CERT_EXPIRING_DAYS = 30
+
+    has_critical = has_major = has_error = False
+
+    applicable_codes = [code for code, ver in fault_min_versions.items() if not cversion.older_than(ver)]
+    if applicable_codes:
+        fault_filter = ",".join('eq(faultInst.code,"{}")'.format(code) for code in applicable_codes)
+        for faultInst in icurl('class', 'faultInst.json?query-target-filter=or({})'.format(fault_filter)):
+            fault_attrs = faultInst['faultInst']['attributes']
+            if fault_attrs['lc'] not in ("raised", "soaking"):
+                continue
+            data.append([fault_attrs['code'], fault_attrs['severity'], fault_attrs['descr']])
+            if fault_attrs['severity'] == 'critical':
+                has_critical = True
+            elif fault_attrs['severity'] == 'major':
+                has_major = True
+
+    if cversion.older_than(FACTORY_CERT_MIN_VERSION) and username and password:
+        date_format = "%b %d %H:%M:%S %Y"
+        current_date_re = re.compile(
+            r'[A-Z][a-z]{2}\s+(?P<mon>[A-Z][a-z]{2})\s+(?P<day>\d+)\s+(?P<time>\d{2}:\d{2}:\d{2})\s+\w+\s+(?P<year>\d{4})'
+        )
+        cert_expiry_re = re.compile(
+            r'notAfter=(?P<date>[A-Z][a-z]{2}\s+\d+\s+\d{2}:\d{2}:\d{2}\s+\d{4})'
+        )
+
+        controllers = (node for node in fabric_nodes if node["fabricNode"]["attributes"]["role"] == "controller")
+        controllers = list(controllers)
+        for controller in controllers:
+            node_attrs = controller["fabricNode"]["attributes"]
+            node_id = node_attrs["id"]
+            node_name = node_attrs["name"]
+            node_addr = node_attrs.get("address")
+            if not node_addr:
+                continue
+            try:
+                c = Connection(node_addr)
+                c.username = username
+                c.password = password
+                c.log = LOG_FILE
+                c.connect()
+                c.cmd("date")
+                current_date_match = current_date_re.search(c.output)
+                c.cmd("acidiag verifyapic")
+                cert_expiry_match = cert_expiry_re.search(c.output)
+            except Exception as e:
+                data.append(["N/A", "error",
+                             "APIC {} ({}): unable to verify factory certificate - {}".format(node_id, node_name, e)])
+                has_error = True
+                continue
+
+            try:
+                current_date = datetime.strptime(
+                    "{mon} {day} {time} {year}".format(**current_date_match.groupdict()), date_format
+                )
+            except (AttributeError, ValueError):
+                data.append(["N/A", "error",
+                             "APIC {} ({}): unable to determine current date".format(node_id, node_name)])
+                has_error = True
+                continue
+
+            try:
+                cert_expiry = datetime.strptime(" ".join(cert_expiry_match.group("date").split()), date_format)
+            except (AttributeError, ValueError):
+                data.append(["N/A", "error",
+                             "APIC {} ({}): unable to determine factory certificate expiry date".format(node_id, node_name)])
+                has_error = True
+                continue
+
+            if cert_expiry <= current_date:
+                data.append(["N/A", "critical",
+                             "APIC {} ({}): factory certificate expired on {} UTC".format(node_id, node_name, cert_expiry)])
+                has_critical = True
+            elif (cert_expiry - current_date).days <= FACTORY_CERT_EXPIRING_DAYS:
+                data.append(["N/A", "major",
+                             "APIC {} ({}): factory certificate expiring on {} UTC".format(node_id, node_name, cert_expiry)])
+                has_major = True
+
+    if data:
+        if has_error:
+            result = ERROR
+            recommended_action = 'Manually verify the factory certificate expiry using `acidiag verifyapic` on the affected APIC(s).'
+        elif has_critical and has_major:
+            result = FAIL_O
+            recommended_action = 'Renew expired certificate(s) immediately. For certificate(s) approaching expiry, renew before they expire to avoid service disruption.'
+        elif has_critical:
+            result = FAIL_O
+            recommended_action = 'Renew the certificate(s) immediately to restore functionality.'
+        elif has_major:
+            result = MANUAL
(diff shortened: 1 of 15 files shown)


asraf-khan:
Attached log for reference.

APIC Log :

[APIC Run log - SSL Cert expiry issue.docx](https://github.com/user-attachments/files/30120149/APIC.Run.log.-.SSL.Cert.expiry.issue.docx)

Pytest Log:

[Pytest log - SSL Cert expiry issue.docx](https://github.com/user-attachments/files/30120157/Pytest.log.-.SSL.Cert.expiry.issue.docx)

asraf-khan:
Attaching latest APIC & Pytest logs

APIC Logs:

[APIC Run log - SSL Cert expiry issue.docx](https://github.com/user-attachments/files/30410627/APIC.Run.log.-.SSL.Cert.expiry.issue.docx)

Pytest Logs:

[Pytest log - SSL Cert expiry issue.docx](https://github.com/user-attachments/files/30410633/Pytest.log.-.SSL.Cert.expiry.issue.docx)

thjonson:
This is great — thanks for adding fault-based cert expiration detection. One
gap I noticed: this approach only surfaces an issue once ACI has already
raised the fault (F4501/F4502/etc.), which means the earliest signal is
whatever threshold ACI's fault logic uses internally, and it doesn't cover
node SSL certs the way `pkiFabricNodeSSLCertificate` does.

As a complementary check (not a replacement), I've been running a script
that queries `pkiFabricNodeSSLCertificate` directly (joined with
`fabricNode` for name/model/serial/role) and computes days-to-expiry per
node, independent of whether a fault has fired yet:

- **EXPIRED** — already past `validityNotAfter`
- **CRITICAL** — expires within ~90 days
- **WARN** — expires within ~365 days (well ahead of any fault threshold)
- Flags certs not chained to "Cisco Manufacturing CA"
- Flags `notAfter` in year 2099, which matches FN-72339 (renewed-SUDI
  signature) — not something the fault codes above catch, since it isn't a
  fault condition, just a suspicious cert attribute worth a manual review.

Since fabric node SSL cert expiry/re-validation failures are a distinct
failure mode from the SAML/KeyRing/TP/Factory certs this PR covers (nodes
stay Inactive on reboot if their SSL cert fails re-validation), would it
make sense to add this as an additional check in this same PR, or should it
be a separate follow-on check? Happy to share the reference script/logic if
useful.

asraf-khan:
@Thjonson Thanks for sharing your suggesstions.


In this PR itself added validation for leaf/spine certificate expiration using MO's "pkiFabricNodeSSLCertificate" irrespective of faults and alert the customer if it's expired and expiring (expires within ~30 days) which matched the days with expired & expiring fault codes and we maintain same with faults expiring & MO's based expiring.

can you please provide details for below item ? how to verify ? what is the impact of this customer will face ? what is the recommended_action we have to provide to customer if they face this ? 

Flags notAfter in year 2099, which matches FN-72339 (renewed-SUDI
signature) — not something the fault codes above catch, since it isn't a
fault condition, just a suspicious cert attribute worth a manual review.

asraf-khan:
@monrog2 this PR still hold validation for APIC Side cert expiry validation. Only leaf/spine cert validation has been reverted.

thjonson:
Quick clarification first: looking at the current diff, I don't see the
leaf/spine `pkiFabricNodeSSLCertificate` check in the PR right now —
`certificate_expiration_check()` currently only queries `faultInst` for
F4501/F4502 (KeyRing), F4503/F4617 (TP), F3081/F3082 (SAML), and F4752/
F4753 (Factory), gated by `fault_min_versions`. I think this lines up with
the "Removing leaf/spine certs check from validation" commit — sounds like
the `pkiFabricNodeSSLCertificate`-based leaf/spine check you described got
reverted before I saw your reply. So just flagging that the gap I raised
(no direct MO polling, and fault-based detection is version-gated to when
each fault code was introduced) still applies to the PR as it stands today
— let me know if that leaf/spine check is coming back in a follow-up.

On to FN-72339 specifically:

I hadn't realized your earlier changes had polled
`pkiFabricNodeSSLCertificate` via moquery directly rather than relying on
the fault codes exclusively — that's a meaningful advantage worth keeping
in mind for a future pass, since it also covers ACI versions **before**
F4501/F4502/etc. were introduced (e.g. pre-3.1(2f) for the SAML fault,
pre-6.0(4c) for KeyRing, etc.), where fault-based detection can't fire
because the fault doesn't exist yet on that code train. Polling the object
directly sidesteps that version-gating problem entirely.

## How to verify FN-72339

FN-72339 hits switches manufactured after June 2020 (dual-cert ACT2 chip)
that get **downgraded** to 4.1(1i)/14.1(1i). On an affected node, `moquery
-c pkiFabricNodeSSLCertificate` shows two tells together:

1. `validityNotAfter` = a date in **2099** (bug CSCvv44023 — the switch
   derives the wrong cert as `server.crt`, which carries a 2099 expiry as
   part of its structure)
2. `message` = a parse error along the lines of "failure to parse the
   subject line"

My script checks for these independently, as flags rather than a single
hard rule, because either one *alone* doesn't necessarily mean broken:

- `!2099` fires when `validityNotAfter` contains "2099"
- `!msg` fires whenever the `message` attribute is non-empty (any parse/
  validation error text)
- `!issuer` fires when the cert issuer doesn't chain to "Cisco
  Manufacturing CA" (a broader integrity check, not FN-72339-specific)

The reason I don't treat a lone `!2099` as an automatic FAIL: a
legitimately-renewed SUDI cert can also carry a far-future `notAfter` as
part of a normal renewal, so 2099 by itself is "worth a look," not
"broken." **`!2099` combined with `!msg`** (or `!issuer`) is the real
FN-72339 signature — that combination is what indicates the switch has an
actual identity/discovery problem, not just an unusual-but-valid cert.

## Recommended action / remediation

- If `!2099` alone (no `!msg`, chains to Cisco Manufacturing CA): informational only, no action required — likely a legitimately renewed cert.
- If `!2099` + `!msg` (and/or `!issuer`): flag as **CRITICAL / do not upgrade or reboot this node yet.** Per FN-72339, the fix is to get the node onto 4.1(1j)/14.1(1j) or later *before* the next reload, then let it do a clean reload so it re-derives the certificate correctly from the ACT2 chip. If the node is already stuck (failed discovery / already rebooted into this state), open a TAC case referencing FN-72339 / CSCvv44023 rather than attempting to force cert regeneration manually.

Let me know if it'd help to see the actual bash logic — it's about 10 lines, happy to paste it inline here.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,122 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+
+    applicable_codes = [code for code, ver in fault_min_versions if not cversion.older_than(ver)]
+    if applicable_codes:
+        fault_filter = ",".join('eq(faultInst.code,"{}")'.format(code) for code in applicable_codes)
+        for faultInst in icurl('class', 'faultInst.json?query-target-filter=or({})'.format(fault_filter)):
+            fault_attrs = faultInst['faultInst']['attributes']
+            if fault_attrs['lc'] not in ("raised", "soaking"):
+                continue
+            data.append([fault_attrs['code'], fault_attrs['severity'], fault_attrs['descr']])
+            if fault_attrs['severity'] == 'critical':
```
Priyanka-Patil14, line 6732: Using severity attribute to classify expired vs expiring is fragile. Better to check the fault code directly against the known expired/expiring sets.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,122 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+                data.append(["N/A", "critical",
+                             "APIC {} ({}): factory certificate expired on {} UTC".format(controller_id, controller_name, cert_expiry)])
+                has_critical = True
+            elif (cert_expiry - current_date).days <= CERT_EXPIRING_DAYS:
+                data.append(["N/A", "major",
+                             "APIC {} ({}): factory certificate expiring on {} UTC".format(controller_id, controller_name, cert_expiry)])
+                has_major = True
+
+    if data:
+        if has_error:
```
Priyanka-Patil14, line 6802: No test covers the case where fault API returns expired certs (has_critical=True) AND SSH to a controller fails (has_error=True) at the same time. In that scenario the code returns ERROR and hides the confirmed FAIL_O finding.

tests/checks/certificate_expiration_check/test_certificate_expiration_check.py
```diff
@@ -0,0 +1,490 @@
+import os
```
Priyanka-Patil14, line 1: No test covers lc: soaking state - the detection spec says soaking faults must also be caught but all fixtures only use raised.

monrog2:
## Integration test results

The certificate expiration check was exercised against the full integration matrix after merging the latest `v4.3.0-dev` changes.

- Python 2.7 unit tests: passed
- Python 3.8 unit tests: passed
- Local full suite: 1,171 passed
- Integration matrix: all seven version scenarios executed
- Date parsing regression: fixed. No `unable to determine current date` results remained in any completed artifact set.

The integration jobs still reported failures from broader runner/fabric connectivity and transient API issues. For the certificate check specifically, the actual-current-version run produced 9 `PASS`, 7 valid `FAIL_O`, and 1 transient API timeout across 17 available fabric artifacts.

Environment-specific names and identifiers have been redacted below.

### Valid `FAIL_O` output

#### Fabric 1

```text
Certificate Expiration Check...  FAIL - OUTAGE WARNING!!
  Fault Code  Description
  ----------  -----------
  F3082       SAML Encryption Certificate has expired
```

#### Fabric 2

```text
Certificate Expiration Check...  FAIL - OUTAGE WARNING!!
  Fault Code  Description
  ----------  -----------
  N/A         APIC [redacted]: factory certificate expired on [redacted] UTC
  N/A         APIC [redacted]: factory certificate expired on [redacted] UTC
  N/A         APIC [redacted]: factory certificate expired on [redacted] UTC
```

#### Fabric 3

```text
Certificate Expiration Check...  FAIL - OUTAGE WARNING!!
  Fault Code  Description
  ----------  -----------
  F4502       KeyRing Certificate [redacted] expired
```

#### Fabric 4

```text
Certificate Expiration Check...  FAIL - OUTAGE WARNING!!
  Fault Code  Description
  ----------  -----------
  F3082       SAML Encryption Certificate has expired
```

#### Fabric 5

```text
Certificate Expiration Check...  FAIL - OUTAGE WARNING!!
  Fault Code  Description
  ----------  -----------
  F4503       TP Certificate [redacted] expired
  F4503       TP Certificate [redacted] expired
```

#### Fabric 6

```text
Certificate Expiration Check...  FAIL - OUTAGE WARNING!!
  Fault Code  Description
  ----------  -----------
  F3082       SAML Encryption Certificate has expired
  F4753       Node [redacted] Factory Certificate expired
  F4753       Node [redacted] Factory Certificate expired
  F4753       Node [redacted] Factory Certificate expired
```

#### Fabric 7

```text
Certificate Expiration Check...  FAIL - OUTAGE WARNING!!
  Fault Code  Description
  ----------  -----------
  F4753       Node [redacted] Factory Certificate expired
  F4753       Node [redacted] Factory Certificate expired
  F4753       Node [redacted] Factory Certificate expired
  F4753       Node [redacted] Factory Certificate expired
  F4753       Node [redacted] Factory Certificate expired
```

All valid findings returned the expected outage-warning action directing the operator to resolve certificate conditions before beginning the upgrade and linking to the certificate-check documentation.

monrog2:
Superseded by #449, which carries the conflict-resolved and integration-tested certificate expiration validation on a repository-owned branch. The replacement PR preserves credit to @asraf-khan and retains the Fixes #408 issue linkage.

PR #418, opened 2026-07-27
Added APIC OOB connectivity checks for CSCwu91693

### LINKED ISSUE #437: NewValidation: CSCwu91693 - APIC OOB connectivity before upgrade

### Validation Type

[ ] - Fault

[x] - Config

[x] - Bug

[ ] - Other

CSCwu91693 describes broken OOB management connectivity between APIC controllers, which can cause a partial or failed cluster upgrade.

The defect is Verified and customer visible. The validation should verify that every APIC has complete, usable OOB configuration and that the required APIC-to-APIC HTTPS paths are reachable before upgrade.

### Expected behavior

- Inspect the complete APIC controller inventory.
- Determine the effective HTTPS port deterministically from the active `commHttps` policy.
- Validate every required APIC-to-APIC OOB path, rather than testing all peers only from the APIC running the script.
- Return `FAIL_UF` when required OOB connectivity is unavailable.
- Return `ERROR` when required controller, OOB address, or HTTPS-policy inventory is missing, malformed, partial, or ambiguous.
- Return `MANUAL` when the target version required for applicability cannot be determined.
- Do not silently skip controllers without usable OOB addresses or convert incomplete evaluation into `PASS` or `N/A`.
- Preserve malformed controller evidence in the reported result.
- Reconcile the target-version gate and fixed-release behavior with the authoritative CSCwu91693 record.

### Testing requirements

Include Python 2.7 and Python 3.8 coverage for:

- All controllers reachable.
- One or more controllers unreachable.
- Empty and partial controller inventory.
- Missing, malformed, link-local-only, and mixed OOB addresses.
- Default, custom, missing, malformed, and multiple `commHttps` policies.
- Missing target version.
- Target-version applicability boundaries.
- Mixed valid and malformed controller objects.
- Effective string-normalized result payloads.

Priyanka-Patil14:
**Summary:**

- This PR adds a new validation check: APIC OOB Connectivity.

- The check detects broken OOB management connectivity between APIC controllers, which can cause a partial or failed cluster upgrade.

**What Changed:**

- Added apic_oob_connectivity_check with nested helper _get_apic_oob_connectivity in aci-preupgrade-validation-script.py
- Added validation documentation in docs/docs/validations.md
- Added dedicated unit tests and test data under: tests/checks/apic_oob_connectivity_check/

**Check Behavior:**

**APIC OOB Connectivity:**

- Returns N/A if target version is below 6.0(2a)
- Returns PASS if all APIC OOB IPs are reachable on the required HTTPS ports
- Returns FAIL_UF if any APIC OOB IP is unreachable
- Returns ERROR if commHttps port cannot be read

**Test Results:**

- Pytest Logs: Full Run:
[CSCwu91693_PytestFullRun_logs.txt](https://github.com/user-attachments/files/30583997/CSCwu91693_PytestFullRun_logs.txt)

- Pytest Logs: apic_oob_connectivity_check:
[CSCwu91693_PytestLogs.txt](https://github.com/user-attachments/files/30584002/CSCwu91693_PytestLogs.txt)

- Script Full Run:
[CSCwu91693_FullRun_logs.txt](https://github.com/user-attachments/files/30512660/CSCwu91693_FullRun_logs.txt)
**Note:** stale_dbgacEpgSummaryTask_check shows an ERROR (name 'timedelta' is not defined) in the full run, this is a pre-existing issue in v4.2.0-dev, not introduced by this PR. It will be resolved by Muthu's separate PR.

- Script Run Logs (PASS / NA / FAIL):
[CSCwu91693_PASS:NA:FAIL_logs.txt](https://github.com/user-attachments/files/30512662/CSCwu91693_PASS.NA.FAIL_logs.txt)

Fixes #437


asraf-khan (changes requested):
Make it as single function check for 2 different validation with help of helper inside same function.

Priyanka-Patil14:
> Make it as single function check for 2 different validation with help of helper inside same function.

Done. Merged both validations into a single check function

docs/docs/validations.md
```diff
@@ -2846,6 +2848,18 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
 Contact Cisco TAC for next steps. For more details, refer to the workaround in [CSCwt69100][75].
 
 
+### APIC OOB Connectivity
+
+Due to [CSCwu91693][77], when an APIC cluster upgrade is triggered, the orchestrating APIC fans out an HTTPS POST to every peer APIC over the OOB management network. If OOB connectivity to any peer APIC is broken at upgrade time, only the reachable APICs receive the trigger and start upgrading. The unreachable APICs are silently skipped, leaving the cluster partially upgraded — a state that cannot be recovered remotely.
```
lovkeshsharma702: Apic bootx use https POST for upgrade starting from 602. Due to [CSCwu91693][77], If OOB connectivity to any peer APIC is broken during upgrade, only the reachable APICs receive the trigger and start upgrading. The unreachable APICs are silently skipped, leaving the cluster partially upgraded.

docs/docs/validations.md
```diff
@@ -2846,6 +2848,18 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
...
+
+This check performs two verifications:
+
+1. **Default port (443)**: Used by APIC bootx starting from 6.0(2). Applicable when target version is 6.0(2) or above.
```
lovkeshsharma702: 1. **Default port (443)**: APIC used bootx starting from 6.0(2). default https port used is 443.  
remote all AI marked special symbol.

docs/docs/validations.md
```diff
@@ -2846,6 +2848,18 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
...
+2. **Custom HTTPS port (from `commHttps`)**: Used by the upgrade fanout starting from 6.2(1). Applicable when both current and target versions are 6.2(1) or above. If the configured port is the same as the default (443), this step is skipped as it is already covered above.
```
lovkeshsharma702: 2. Custom HTTPS port (from `commHttps`): starting from 6.2(1) Customer ports are supported. Applicable when both current and target versions are 6.2(1) or above.

docs/docs/validations.md
```diff
@@ -2846,6 +2848,18 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
...
+
+For each applicable check, the script queries `topSystem` filtered to `role=controller` to collect all APIC OOB management IPs and attempts a TCP connection on the relevant port with a 5-second timeout. If any peer is unreachable, the check fails.
```
lovkeshsharma702: we are using other fuction also to get OOB. please use these.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
     return Result(result=result, headers=headers, data=data, recommended_action=recommended_action, doc_url=doc_url)
 
 
+@check_wrapper(check_title="APIC OOB Connectivity")
```
lovkeshsharma702: utilise previously collected data in script. mgmtRsOoBStNode,  mgmtOoBs = icurl("class", "mgmtOoB.json?rsp-subtree=children") or so on.

lovkeshsharma702 (changes requested):

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+        data = []
+        has_error = False
+
+        for topSystem in topSystems:
+            attrs = topSystem['topSystem']['attributes']
+            node_id = attrs.get('id', '')
+
+            if attrs.get('oobMgmtAddr', '0.0.0.0') != '0.0.0.0':
+                ip = attrs.get('oobMgmtAddr')
+            elif attrs.get('oobMgmtAddr6', '::') not in ('', '::', '0:0:0:0:0:0:0:0'):
```
lovkeshsharma702: get node id from FabricNode and previously called OOB fuctions and use that.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+    headers = ["Node ID", "OOB IP", "Port", "Status"]
+    recommended_action = "Restore OOB management connectivity between all APICs and ensure the required HTTPS ports are reachable across the OOB network."
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#apic-oob-connectivity'
+
+    def _get_apic_oob_connectivity(topSystems, port):
+        data = []
+        has_error = False
+
+        for topSystem in topSystems:
+            attrs = topSystem['topSystem']['attributes']
```
lovkeshsharma702: getting node id and ip for all nodes is bad idea. get it only for apic. /api/node/class/topSystem.json?query-target-filter=and(eq(topSystem.role,"controller")). please used previously collected Data in old fuctions.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+            elif attrs.get('oobMgmtAddr6', '::') not in ('', '::', '0:0:0:0:0:0:0:0'):
+                ip = attrs.get('oobMgmtAddr6')
+            else:
+                continue
+
+            try:
+                if subprocess.call(
+                    'curl --max-time 5 -k -s -o /dev/null https://{}:{} 2>/dev/null'.format(ip, port),
+                    shell=True
+                ) in [7, 28]:
```
lovkeshsharma702: anything as !0 should be tracked as failure.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+    has_error = False
+
+    # Default port check: APIC bootx uses port 443 from 6.0(2)
+    default_data, default_error = _get_apic_oob_connectivity(topSystems, 443)
+    data.extend(default_data)
+    if default_error:
+        has_error = True
+
+    # Custom HTTPS port check: upgrade fanout uses commHttps port from 6.2(1)
+    if not (cversion.older_than("6.2(1a)") or tversion.older_than("6.2(1a)")):
```
lovkeshsharma702: if (cversion.newer_than("6.2(1a)"). target version can be ignored.

lovkeshsharma702 (changes requested):

docs/docs/validations.md
Priyanka-Patil14: As discussed, keeping topSystem approach.

aci-preupgrade-validation-script.py
Priyanka-Patil14: As discussed, keeping topSystem approach

aci-preupgrade-validation-script.py
Priyanka-Patil14: As discussed, keeping topSystem approach

aci-preupgrade-validation-script.py
Priyanka-Patil14: As discussed, keeping topSystem approach.

docs/docs/validations.md
Priyanka-Patil14: Updated.

docs/docs/validations.md
Priyanka-Patil14: Updated.

aci-preupgrade-validation-script.py
Priyanka-Patil14: Updated

aci-preupgrade-validation-script.py
Priyanka-Patil14: Updated to if not cversion.older_than "6.2(1a)".

docs/docs/validations.md
Priyanka-Patil14: Updated.

docs/docs/validations.md
```diff
@@ -2846,6 +2848,18 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
...
+### APIC OOB Connectivity
+
+APIC bootx uses HTTPS POST for upgrade starting from 6.0(2). Due to [CSCwu91693][77], if OOB connectivity to any peer APIC is broken during upgrade, only the reachable APICs receive the trigger and start upgrading. The unreachable APICs are silently skipped, leaving the cluster partially upgraded.
+
+This check performs two verifications:
+
+1. **Default port (443)**: APIC uses bootx starting from 6.0(2). Default HTTPS port used is 443.
+2. **Custom HTTPS port (from `commHttps`)**: Starting from 6.2(1), custom ports are supported. Applicable when current version is 6.2(1) or above.
+
+For each applicable check, the script queries `topSystem` filtered to `role=controller` to collect all APIC OOB management IPs and attempts a TCP connection on the relevant port with a 5-second timeout. If any peer is unreachable, the check fails.
```
lovkeshsharma702: ### APIC OOB Connectivity

"Starting from 6.0(2), APIC firmware upgrades are triggered via an HTTPS POST request (bootx) sent to each peer APIC over its out-of-band (OOB) management interface. Due to [CSCwu91693][77], if OOB connectivity to a peer APIC is unavailable at the time this trigger is sent, that APIC does not receive it and silently fails to start the upgrade, while the remaining reachable APICs proceed normally. This results in a partially upgraded cluster with no explicit error raised at the time of failure.

This check verifies OOB reachability between APICs on the port(s) actually used for the upgrade trigger. The default port 443 is validated on all versions from 6.0(2) onward, since it is always used unless a custom HTTPS port is configured. From 6.2(1) onward, the upgrade trigger also honors a custom HTTPS port if one is configured via the `commHttps` policy; this custom port is validated only when both the current and target versions are 6.2(1) or later, and only when the configured port differs from 443, which is already covered by the default check.

For each applicable port, the script queries `topSystem` filtered to `role=controller` to collect the OOB management IP of every APIC in the cluster, then attempts an HTTPS connection to each peer on that port with a 5-second timeout. If any APIC is found unreachable on a port used for the upgrade trigger, the check fails."

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+            elif attrs.get('oobMgmtAddr6', '::') not in ('', '::', '0:0:0:0:0:0:0:0'):
+                ip = attrs.get('oobMgmtAddr6')
+            else:
+                continue
+
+            try:
+                if subprocess.call(
+                    'curl --max-time 5 -k -s -o /dev/null https://{}:{} 2>/dev/null'.format(ip, port),
+                    shell=True
+                ) not in [0, 22]:
```
lovkeshsharma702: keep exit code ok as 0 and raise alarm for any other code.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+    has_error = False
+
+    # Default port check: APIC bootx uses port 443 from 6.0(2)
+    default_data, default_error = _get_apic_oob_connectivity(topSystems, 443)
+    data.extend(default_data)
+    if default_error:
+        has_error = True
+
+    # Custom HTTPS port check: upgrade fanout uses commHttps port from 6.2(1)
+    if not cversion.older_than("6.2(1a)"):
```
lovkeshsharma702: use only CCO version in all checks

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+                port = int(commHttps[0]['commHttps']['attributes'].get('port', 443))
+            except (ValueError, KeyError):
+                log.warning("Could not read commHttps port")
+                return Result(result=ERROR, msg="Could not read commHttps HTTPS port from commHttps MO.")
+
+        if port != 443:  # Port 443 already covered by default port check above
+            custom_data, custom_error = _get_apic_oob_connectivity(topSystems, port)
+            data.extend(custom_data)
+            if custom_error:
+                has_error = True
```
lovkeshsharma702: _  remove _before starting the fuction.

lovkeshsharma702 (changes requested):

aci-preupgrade-validation-script.py
Priyanka-Patil14: Updated.

aci-preupgrade-validation-script.py
Priyanka-Patil14: Used CCO version.

aci-preupgrade-validation-script.py
Priyanka-Patil14: Removed.

docs/docs/validations.md
Priyanka-Patil14: Updated.

tests/checks/apic_oob_connectivity_check/topSystem_3apics_oob.json
```diff
@@ -0,0 +1,32 @@
+[
```
lovkeshsharma702, line 1: missing testcase for both ipv6. please add that as well and test in lab.

lovkeshsharma702 (changes requested):

tests/checks/apic_oob_connectivity_check/topSystem_3apics_oob.json
Priyanka-Patil14: Added.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+
+        return data, has_error
+
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+
+    if tversion.older_than("6.0(2a)"):
+        return Result(result=NA, msg=VER_NOT_AFFECTED)
+
+    topSystems = icurl('class', 'topSystem.json?query-target-filter=eq(topSystem.role,"controller")')
```
lovkeshsharma702: topSystems as variable is confusing. please change this to something Apic_id_ip or so .

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+    if not topSystems:
+        return Result(result=NA, msg="No APIC controller nodes found.")
+
+    data = []
+    has_error = False
+
+    # Default port check: APIC bootx uses port 443 from 6.0(2)
+    default_data, default_error = get_apic_oob_connectivity(topSystems, 443)
```
lovkeshsharma702: same topsystem as suggested earlier. Keep it more of variable not mo name.

aci-preupgrade-validation-script.py
lovkeshsharma702: APIC OOB Connectivity --> APIC OOB Connectivity check

tests/checks/apic_oob_connectivity_check/topSystem_3apics_oob.json
lovkeshsharma702, line 1: add more check for having ipv6 and ipv4 combination and see what system choose. in any case it should use ip4 if ipv6 is present also.

lovkeshsharma702 (changes requested):

tests/checks/apic_oob_connectivity_check/topSystem_3apics_oob.json
Priyanka-Patil14: Added.

docs/docs/validations.md
```diff
@@ -2846,6 +2848,15 @@ Affected versions: 6.1(5e) and below, or 6.2(1g).
...
+
+This check verifies OOB reachability between APICs on the port(s) actually used for the upgrade trigger. The default port 443 is validated on all versions from 6.0(2) onward, since it is always used unless a custom HTTPS port is configured. From 6.2(1) onward, the upgrade trigger also honors a custom HTTPS port if one is configured via the `commHttps` policy; this custom port is validated only when the current version is 6.2(1) or later, and only when the configured port differs from 443, which is already covered by the default check.
+
+For each applicable port, the script queries `topSystem` filtered to `role=controller` to collect the OOB management IP of every APIC in the cluster, then attempts an HTTPS connection to each peer on that port with a 5-second timeout. If any APIC is found unreachable on a port used for the upgrade trigger, the check fails.
```
lovkeshsharma702: For each applicable port, the script get OOB management IP of every APIC in the cluster, then attempts an HTTPS connection to each peer on that port with a 5-second timeout. If any APIC is found unreachable on a port used for the upgrade trigger, the check fails.

docs/docs/validations.md
```diff
@@ -206,6 +206,7 @@ Items                                           | Defect       | This Script
 [WRED with Affected FM Models][d35]             | CSCwt50713   | :white_check_mark: | :no_entry_sign:
 [N9K-C93180YC-FX3 Switch Memory Less Than 32GB][d36] | CSCwm42741   | :white_check_mark: | :no_entry_sign:
 [Stale dbgacEpgSummaryTask Objects][d37]         | CSCwt69100   | :white_check_mark: | :no_entry_sign:
+[APIC OOB Connectivity][d38]                      | CSCwu91693   | :white_check_mark: | :white_check_mark: 6.2(2)
```
lovkeshsharma702: [APIC OOB Connectivity][d38]                      | CSCwu91693   | :white_check_mark: | :white_check_mark: 6.2(2) 
---> correct.
[APIC OOB Connectivity][d38]                      | CSCwu91693   | :white_check_mark: |  :no_entry_sign:

lovkeshsharma702 (changes requested):

aci-preupgrade-validation-script.py
Priyanka-Patil14: Updated.

aci-preupgrade-validation-script.py
Priyanka-Patil14: Updated.

docs/docs/validations.md
Priyanka-Patil14: Updated.

docs/docs/validations.md
Priyanka-Patil14: Corrected.

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,85 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+
+        for apic in apic_id_ip:
+            attrs = apic['topSystem']['attributes']
+            node_id = attrs.get('id', '')
+
+            if attrs.get('oobMgmtAddr', '0.0.0.0') != '0.0.0.0':
+                ip = attrs.get('oobMgmtAddr')
+            elif attrs.get('oobMgmtAddr6', '::') not in ('', '::', '0:0:0:0:0:0:0:0'):
+                ip = attrs.get('oobMgmtAddr6')
+            else:
```
lovkeshsharma702, line 6724: curl --max-time 5 -k -s https://[2001:db8:abc:1::12]:443 -- make sure you use proper format for ipv6.  please check why pytest is noyt handling this.

lovkeshsharma702 (changes requested):

aci-preupgrade-validation-script.py
Priyanka-Patil14: Updated.

monrog2 (changes requested):
Good implementation overall. A few suggestions to strengthen test coverage and version boundary validation. See inline comments.

tests/checks/apic_oob_connectivity_check/test_apic_oob_connectivity_check.py
```diff
@@ -0,0 +1,185 @@
...
+import importlib
+from helpers.utils import read_data
+
+script = importlib.import_module("aci-preupgrade-validation-script")
+
+log = logging.getLogger(__name__)
+dir = os.path.dirname(os.path.abspath(__file__))
+
+test_function = "apic_oob_connectivity_check"
+
```
monrog2, line 13: **[Medium] Missing version boundary test immediately below 6.0(2a)**

The test suite covers `tversion="5.2(7f)"` (well below) and `tversion="6.0(2a)"` (at boundary), but does not test immediately below the boundary (e.g., `6.0(1h)`).

Per project review guidelines, version gate checks should test "immediately below, at, within, and immediately above each boundary" to catch off-by-one errors in version comparison logic.

**Suggested addition:**
```python
# tversion = 6.0(1h) (immediately below 6.0(2a)) -> NA
(
    {topSystem: [], commHttps: []},
    "6.0(1h)",
    "6.0(1h)",
    [],
    script.NA,
),
```

tests/checks/apic_oob_connectivity_check/test_apic_oob_connectivity_check.py
```diff
@@ -0,0 +1,185 @@
...
+        # tversion >= 6.0(2a), all APICs have no OOB configured -> PASS
+        (
+            {topSystem: read_data(dir, "topSystem_no_oob.json"), commHttps: []},
+            "6.0(2a)",
+            "6.0(3a)",
+            [],
+            script.PASS,
+        ),
+        # tversion >= 6.0(2a), all APICs reachable on port 443 -> PASS
+        (
```
monrog2, line 55: **[Medium] Missing cversion boundary test immediately below 6.2(1g)**

The check has a version gate `if not cversion.older_than("6.2(1g)")` that enables custom port checking. Tests cover `cversion="6.2(1g)"` but don't test `cversion="6.2(1f)"` (immediately below) to confirm custom port logic is correctly skipped.

**Suggested addition:**
```python
# cversion = 6.2(1f) (immediately below 6.2(1g)), custom port configured -> PASS
# (custom port check should be SKIPPED, only default 443 check runs)
(
    {
        topSystem: read_data(dir, "topSystem_3apics_oob.json"),
        commHttps: read_data(dir, "commHttps_custom_port.json"),
    },
    "6.2(1f)",
    "6.2(2a)",
    [0, 0, 0],  # Only 3 curl calls for port 443, not 6
    script.PASS,
),
```

tests/checks/apic_oob_connectivity_check/test_apic_oob_connectivity_check.py
```diff
@@ -0,0 +1,185 @@
...
+        ),
+    ],
+)
+def test_logic(run_check, mock_icurl, monkeypatch, cversion, tversion, curl_exit_codes, expected_result):
+    idx = [0]
+
+    def mock_subprocess_call(cmd, shell=False):
+        # Verify that IPv6 addresses in the curl URL are wrapped in square brackets
+        import re
+        match = re.search(r'https://([^/]+):', cmd)
```
monrog2, line 168: **[Medium] Tests only assert result status, not payload content**

All tests only verify `result.result == expected_result`. None verify `result.data`, `result.headers`, `result.msg`, or `result.recommended_action`.

Per project review guidelines: *"Require status and payload assertions, not only execution without exception."*

A check could return `FAIL_UF` with incorrect or empty data, misleading users about which APICs are unreachable.

**Suggested addition for FAIL_UF cases:**
```python
assert result.result == expected_result
if expected_result == script.FAIL_UF:
    assert result.headers == ["Node ID", "OOB IP", "Port", "Status"]
    assert len(result.data) > 0, "FAIL_UF should include affected APIC data"
    # Verify at least one row shows 'Unreachable'
    assert any(row[3] == "Unreachable" for row in result.data)
```

aci-preupgrade-validation-script.py
```diff
@@ -6702,6 +6702,86 @@ def stale_dbgacEpgSummaryTask_check(tversion, **kwargs):
...
+                continue
+
+            try:
+                ip_formatted = '[{}]'.format(ip) if ':' in ip else ip
+                if subprocess.call(
+                    'curl --max-time 5 -k -s -o /dev/null https://{}:{} 2>/dev/null'.format(ip_formatted, port),
+                    shell=True
+                ) != 0:
+                    data.append([node_id, ip, port, "Unreachable"])
+            except Exception as e:
```
monrog2, line 6734: **[Low] Consider avoiding `shell=True` in subprocess call**

Using `shell=True` with string formatting is generally discouraged. While not a security risk here (IPs come from trusted APIC API), using a list is cleaner and more portable:

```python
if subprocess.call(
    ['curl', '--max-time', '5', '-k', '-s', '-o', '/dev/null',
     'https://{}:{}'.format(ip_formatted, port)],
    stderr=subprocess.DEVNULL
) != 0:
```

This is a minor suggestion and not blocking.

monrog2 (changes requested):
Good implementation overall with proper Result semantics and version gating. A few suggestions to strengthen test coverage and version boundary validation.

lovkeshsharma702 (approved):
all comments have been addressed.

aci-preupgrade-validation-script.py
```diff
@@ -6797,6 +6797,88 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+                data.append([node_id, ip, port, "Error"])
+                has_error = True
+                continue
+
+        return data, has_error
+
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+
+    if tversion.older_than("6.0(2a)"):
```
monrog2, line 6842: **[Blocking] Reconcile this applicability gate with the defect contract.**

The check definition says `tversion >= 6.0(2)`, while the CSCwu91693 RNE condition says the *running APIC* is 6.0.2 or later. Those differ when an upgrade crosses 6.0(2). Please obtain defect-owner confirmation, then align the implementation and documentation to one explicit current/target matrix. Add cases for current below 6.0(2) with target at/above it, and current at/above 6.0(2).

aci-preupgrade-validation-script.py
```diff
@@ -6797,6 +6797,88 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+        for apic in apic_id_ip:
+            attrs = apic['topSystem']['attributes']
+            node_id = attrs.get('id', '')
+
+            if attrs.get('oobMgmtAddr', '0.0.0.0') != '0.0.0.0':
+                ip = attrs.get('oobMgmtAddr')
+            elif attrs.get('oobMgmtAddr6', '::') not in ('', '::', '0:0:0:0:0:0:0:0'):
+                ip = attrs.get('oobMgmtAddr6')
+            else:
+                continue
```
monrog2, line 6820: **[Blocking] Do not skip a controller that has no usable OOB address.**

Skipping it allows the new upgrade-blocking validation to return `PASS` without checking that peer. Return `FAIL_UF` with node-level evidence such as `OOB address not configured`. Also reject link-local-only IPv6 unless the required scope/interface is available: on the live 6.0(8e APIC, `topSystem.oobMgmtAddr6` contained `fe80::` addresses while `mgmtRsOoBStNode.v6Addr` was `::`. Please cover no-address, link-local-only, IPv4-only, and configured global-IPv6 cases.

aci-preupgrade-validation-script.py
```diff
@@ -6797,6 +6797,88 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+
+    if not tversion:
+        return Result(result=MANUAL, msg=TVER_MISSING)
+
+    if tversion.older_than("6.0(2a)"):
+        return Result(result=NA, msg=VER_NOT_AFFECTED)
+
+    apic_id_ip = icurl('class', 'topSystem.json?query-target-filter=eq(topSystem.role,"controller")')
+    if not apic_id_ip:
+        return Result(result=NA, msg="No APIC controller nodes found.")
```
monrog2, line 6847: **[Blocking] Empty or incomplete required inventory must fail closed.**

`N/A` becomes a passed, hidden validation, but a running APIC cannot legitimately establish this check with zero controllers. Return `ERROR` for an empty response. Also compare the returned controller IDs with the controller inventory already available to the script so a partial `topSystem` response cannot produce `PASS`. Add empty and partial inventory tests.

aci-preupgrade-validation-script.py
```diff
@@ -6797,6 +6797,88 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+    has_error = False
+
+    # Default port check: APIC bootx uses port 443 from 6.0(2)
+    default_data, default_error = get_apic_oob_connectivity(apic_id_ip, 443)
+    data.extend(default_data)
+    if default_error:
+        has_error = True
+
+    # Custom HTTPS port check: upgrade fanout uses commHttps port from 6.2(1)
+    if not cversion.older_than("6.2(1g)"):
```
monrog2, line 6859: **[Blocking] Enforce both custom-port version conditions.**

The defined upgrade-fanout logic applies the custom HTTPS port only when both current and target releases are at or above the applicable 6.2 boundary. This branch checks only `cversion`, so a 6.2 current release targeting below the boundary still probes the custom port. Please add the target-version condition and a complete current/target boundary matrix using the defect-owner-confirmed CCO version.

aci-preupgrade-validation-script.py
```diff
@@ -6797,6 +6797,88 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+        port = 443
+        commHttps = icurl('class', 'commHttps.json')
```
monrog2, line 6861: **[Blocking] Resolve the effective `commHttps` policy deterministically and fail closed.**

A class query may return default and non-default communication policies, so selecting `[0]` relies on undefined response order. Hard-coding `comm-default` is not correct either because a non-default policy may be active. Determine the policy actually used by upgrade fanout through the authoritative APIC configuration/relationship, then read that policy's `commHttps` child. Return `ERROR` when the effective policy or port is missing, malformed, or ambiguous. Add fixtures where default and non-default policies coexist and the non-default policy is effective. The live APIC had only `comm-default`, so this selection behavior still requires validation on an appropriate fabric.

aci-preupgrade-validation-script.py
```diff
@@ -6797,6 +6797,88 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+    headers = ["Node ID", "OOB IP", "Port", "Status"]
+    recommended_action = "Restore OOB management connectivity between all APICs and ensure the required HTTPS ports are reachable across the OOB network."
+    doc_url = 'https://datacenter.github.io/ACI-Pre-Upgrade-Validation-Script/validations/#apic-oob-connectivity'
+
+    def get_apic_oob_connectivity(apic_id_ip, port):
+        data = []
+        has_error = False
+
+        for apic in apic_id_ip:
+            attrs = apic['topSystem']['attributes']
```
monrog2, line 6812: **[High] Parse controller objects independently so malformed inventory does not erase valid evidence.**

Direct indexing here lets one malformed object escape to the wrapper, which replaces the whole result with a generic `ERROR` and discards any unreachable controllers already collected. Validate each object's shape, preserve valid failure rows, and report malformed entries through `unformatted_data` or an error indication. Add a mixed valid-plus-malformed fixture.

aci-preupgrade-validation-script.py
```diff
@@ -6797,6 +6797,88 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+                ip = attrs.get('oobMgmtAddr')
+            elif attrs.get('oobMgmtAddr6', '::') not in ('', '::', '0:0:0:0:0:0:0:0'):
+                ip = attrs.get('oobMgmtAddr6')
+            else:
+                continue
+
+            try:
+                ip_formatted = '[{}]'.format(ip) if ':' in ip else ip
+                with open(os.devnull, 'wb') as devnull:
+                    if subprocess.call(
```
monrog2, line 6825: **[Blocking] Verify that this probe topology satisfies APIC-to-APIC reachability.**

This runs every curl from only the APIC executing the script. That proves one source can reach each OOB address, not that every upgrade-fanout source can reach every peer. Please document and validate the product guarantee that the executing APIC is the sole relevant fanout source, or perform the supported remote checks needed to cover every required source-to-peer path. Without that guarantee, the check can pass while another APIC-to-APIC path is broken.

tests/checks/apic_oob_connectivity_check/test_apic_oob_connectivity_check.py
```diff
@@ -0,0 +1,249 @@
...
+        url = cmd[-1]
+        match = re.search(r'https://([^/]+):', url)
+        if match:
+            host = match.group(1)
+            # If host contains a colon (IPv6) it must be wrapped in brackets
+            if ':' in host:
+                assert host.startswith('[') and host.endswith(']'), (
+                    "IPv6 address in curl URL must be wrapped in square brackets, got: {}".format(url)
+                )
+        code = curl_exit_codes[idx[0]] if idx[0] < len(curl_exit_codes) else 0
```
monrog2, line 238: **[Medium] Make unexpected curl calls fail the test.**

Once the supplied exit-code list is exhausted, this mock returns success forever. The below-boundary custom-port test would therefore still pass if production made extra 8443 calls. Fail when `idx` exceeds the sequence and assert the final call count and exact URLs/ports for each version-gate case.

aci-preupgrade-validation-script.py
```diff
@@ -6797,6 +6797,88 @@ def infravlan_overlap_access_policy_check(tversion, **kwargs):
...
+                        ['curl', '--max-time', '5', '-k', '-s', '-o', os.devnull,
+                         'https://{}:{}'.format(ip_formatted, port)],
+                        stderr=devnull
+                    ) != 0:
+                        data.append([node_id, ip, port, "Unreachable"])
```
monrog2, line 6830: **[Low] Keep result-table payload values string-normalized.**

Append `str(port)` rather than an integer and update the payload assertions. This keeps the new rows consistent across sorting, terminal formatting, and JSON consumers.

monrog2 (changes requested):
AI-ACIWISE classifies CSCwu91693 as BEHAVIOUR-CHANGE YES with high confidence across all three models. This PR therefore introduces a new catastrophic upgrade-blocking prerequisite: conditions that prevent complete evaluation must fail closed rather than yield PASS or N/A.

Read-only live validation on APIC 6.0(8e) confirmed that filtered topSystem and fabricNode return matching controller inventories, topSystem exposes OOB IPv4/IPv6 fields, and commHttps exposes the HTTPS port. It also showed that topSystem may expose fe80:: link-local IPv6 while the configured OOB relation has v6Addr=::, so link-local values are not sufficient evidence of usable IPv6 OOB connectivity. The live fabric has only comm-default and cannot establish how an active non-default communication policy is selected.

FAIL_UF is the appropriate result for both default- and custom-port failures because either condition must block the upgrade. Please align the external check specification with that behavior. Keep the conservative any-nonzero-curl-code failure rule; restricting it to only exit codes 7 and 28 could accept other unusable HTTPS paths.

The focused tests and CheckManager/AciResult tests pass, as do Python 2.7/3.8 CI, but the false-pass and policy-selection paths below remain blocking.

monrog2:
Also, what open issue is this closing? please file one if non existing and link them

aci-preupgrade-validation-script.py
lovkeshsharma702: resolved.

